PR #4, opened 2018-03-23
Use ~/.rustup/tmp/ instead of $TMPDIR as the temporary storage

### LINKED ISSUE #2: Download/Install fails moving from tempdir to destination
On Unix system, it is very customary for the temporary directory (/tmp) to not live in the same filesystem as the home directory. So the `fs::rename` in `Toolchain::install` fails with "Invalid cross-device link (os error 18)".

kennytm:
Fix #2.

r? @glandium

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -208,6 +208,7 @@ impl fmt::Display for Toolchain {
 #[derive(Clone, Debug)]
 struct DownloadParams {
     url_prefix: String,
+    tmp_dir: PathBuf,
     install_dir: PathBuf,
     install_cargo: bool,
     force_install: bool,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
So this seems fine, but maybe we should use a different location for `TempDir`? We could create it somewhere in `~/.rustup` which would avoid this problem entirely (and probably make the tool faster, generally speaking).

This looks fine though at least as a stop-gap.

glandium:
`Could not move tempdir into destination: the source path is not an existing regular file`

kennytm:
@Mark-Simulacrum Yes, looks like using `~/.rustup/tmp/` (which already exists) is a better solution.

kennytm:
r? @glandium again. 

Now we use `~/.rustup/tmp` to store the temporary files as suggested in https://github.com/rust-lang-nursery/cargo-bisect-rustc/pull/4#issuecomment-375497759.

glandium:
This works properly. You may also want to add something that cleans up those files when bisect-rustc is interrupted with ctrl-c.

kennytm:
Thanks for checking @glandium 🙂. Installing a signal handler for Ctrl+C would be a big feature, please file another issue for this.

Issue #3, opened 2018-03-22
--start / --end don't seem to work


glandium:
I tried passing various dates for --start/--end, but all I got was a search between the last two nightlies ("nightly-2018-03-21 through nightly-2018-03-22" as of writing).

Relatedly, it would be nice to be able to pass rust versions instead of nightly dates.

kennytm:
@glandium Could you show some example inputs?

glandium:
I've tried `--start nightly-2018-01-01`, `--start 2018-01-01`, and other dates.

kennytm:
The correct format should be `--start 2018-01-01`. I'm checking why `--start 2018-01-01` won't work.

https://github.com/rust-lang-nursery/cargo-bisect-rustc/blob/328524bc615bd44de8187ab3de5ddb3357bd6816/src/main.rs#L130-L138

glandium:
That function appears to return the right value for my input.

glandium:
It looks like the `bisect_nightlies` code completely ignores the arguments.

```rust
   let mut nightly_date = today;
   let mut jump_length = 1;
   // before this date we didn't have -std packages
   let end_at = chrono::Date::from_utc(
       chrono::naive::NaiveDate::from_ymd(2015, 10, 20),
       chrono::Utc,
   );
```

PR #5, opened 2018-03-23
Use the --start and --end arguments when bisecting nightlies

glandium:
Fixes: #3

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -647,7 +647,12 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> Result<BisectionResult, Er
     let dl_spec = DownloadParams::for_nightly(&cfg);
     let now = chrono::Utc::now();
     let today = now.date();
-    let mut nightly_date = today;
+    let (mut nightly_date, has_start) = if let Some(Bound::Date(date)) = cfg.args.start {
+        (date, true)
+    } else {
+        (today, false)
+    };
+
     let mut jump_length = 1;
     // before this date we didn't have -std packages
     let end_at = chrono::Date::from_utc(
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -647,13 +647,22 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> Result<BisectionResult, Er
     let dl_spec = DownloadParams::for_nightly(&cfg);
     let now = chrono::Utc::now();
     let today = now.date();
-    let mut nightly_date = today;
+    let mut nightly_date = if let Some(Bound::Date(date)) = cfg.args.start {
```
Mark-Simulacrum: I think these are reverse from what I would expect (or at least inconsistent with `bisect_ci`)... `start` is the "earliest" point and `end` is the "latest" point as implemented there, so this should be `cfg.args.end` and `end_at` should be `cfg.args.start`. Maybe we should switch them though...

kennytm:
I don't think you should be modifying `end_at`. The algorithm of this function is,

1. **The `while` loop.** Find the nightly at 1, 2, 4, 8, 16, ... days before today, until the test passes. The nightly that passes the test (e.g. T-64) is counted as `first_success`, and the nightly that fails the test (e.g. T-32) is counted as `last_failure`.

2. **The `least_satisfying` call.** We bisect between `first_success .. last_failure` to find the actual breaking commit.

If `--start` and `--end` are both provided, you could skip step 1 and set `first_success = start` and `last_failure = end` directly.

src/main.rs
glandium: --help says the end point should pass the test and the start point should *not* pass the test. That would be the opposite of what you're saying bisect_ci is doing.

Mark-Simulacrum: I don't think so? The start point should compile, therefore not passing the test, whereas the end point shouldn't compile therefore passing the test. I think @kennytm's comment (https://github.com/rust-lang-nursery/cargo-bisect-rustc/pull/5#issuecomment-375505005) also supports my reasoning, though I'm not 100% sure.

(it might be easier to discuss this on IRC, feel free to ping me if you want in #rust-infra as simulacrum).

PR #13, opened 2018-05-08
Add interactive prompt mode

### LINKED ISSUE #9: Add an interactive mode
In some cases, like performance regression, it's hard to write reliable automated tests to drive the bisection. It would be great to have a git-bisect like interactive mode allowing to run commands manually and then tag whether the tested version was good or bad.

anp:
If a user passes the `--prompt` CLI flag, this allows them to manually inspect the stdio and exit status of each test run before deciding whether to mark it as a regression or not. They can also retry the test.

depends on #5
closes #9


Mark-Simulacrum:
Thanks!

PR #19, opened 2018-07-06
Add --script to specify an arbitrary script.

### LINKED ISSUE #11: Option to use a script?
Is there an option to use a script like the old `bisect-rust`?

My use case was to test an incremental regression.  I had a simple one-line shell script which ran `cargo` twice, checking for specific behavior.  I don't see a way to do that with this tool.

ehuss:
Fixes #11

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -138,6 +138,9 @@ struct Opts {
 
     #[structopt(long = "force-install", help = "force installation over existing artifacts")]
     force_install: bool,
+
+    #[structopt(long = "script", help = "script to run instead of cargo to test for regression")]
+    script: Option<String>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


ehuss:
This can help with checking for regressions besides running cargo.

A silly example would be:

test.sh:
```
#!/usr/bin/env bash

ver=`rustc -V`
echo $ver
pat="20[0-9]{2}-[0-9]{2}-[0-9]{2}"
[[ $ver =~ $pat ]]
if [[ "${BASH_REMATCH[0]}" > "2018-06-15" ]]
then
    echo "Failed"
    exit 1
fi
```

`cargo-bisect-rustc --test-dir . --script=./test.sh -vv --preserve --start=2018-06-02 --end=2018-06-29`

src/main.rs
```diff
@@ -138,6 +138,9 @@ struct Opts {
 
     #[structopt(long = "force-install", help = "force installation over existing artifacts")]
     force_install: bool,
+
+    #[structopt(long = "script", help = "script to run instead of cargo to test for regression")]
+    script: Option<String>,
```
kennytm: Could we make this a PathBuf though?

Mark-Simulacrum:
Thanks!

PR #37, opened 2019-10-24
This make the last failure be the default installed nightly if any

spastorino:


src/main.rs
```diff
@@ -464,10 +466,46 @@ enum TestOutcome {
 }
 
 impl Toolchain {
+    fn default_nightly() -> Option<Date<Utc>> {
```
spastorino: We could cache calls to this function :)

spastorino:
I haven't properly tested this :), so let's consider it WIP for now.

src/main.rs
```diff
@@ -464,10 +466,46 @@ enum TestOutcome {
...
+                    let month = cap.get(2)?.as_str().parse::<u32>().ok()?;
+                    let day = cap.get(3)?.as_str().parse::<u32>().ok()?;
+
+                    return Some(Date::from_utc(
+                        naive::NaiveDate::from_ymd(year, month, day),
+                        Utc,
+                    ));
+                }
+            }
+        }
```
Mark-Simulacrum: Let's maybe try to use `rustc-version` crate here? It seems to take care of the low-level details here for us

src/main.rs
Mark-Simulacrum: Can we add a doc comment on this? Something like

This returns the date of the default toolchain, if it is a nightly toolchain. Returns None if the installed toolchain is not a nightly toolchain.

src/main.rs
```diff
@@ -464,10 +466,46 @@ enum TestOutcome {
...
+
+        None
+    }
+
+    fn is_default_nightly(&self) -> bool {
```
Mark-Simulacrum: Maybe `is_current_nightly`?

src/main.rs
```diff
@@ -847,8 +890,18 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> Result<BisectionResult, Er
     let mut last_failure = if let Some(Bound::Date(date)) = cfg.args.end {
         date
     } else {
-        today
+        if let Some(date) = Toolchain::default_nightly() {
+            // start date must be prior to end date
```
Mark-Simulacrum: Could we print something like "Using {date} from the currently installed default nightly toolchain as the last failure" or something like that?

src/main.rs
```diff
@@ -869,10 +922,24 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> Result<BisectionResult, Er
...
+                let jump_length = if current_jump < 7 {
+                    current_jump += 2;
+                    2
+                // from 2nd to 7th week jump weekly
+                } else if current_jump < 49 {
+                    current_jump += 7;
+                    7
+                // from 7th week jump by two weeks
+                } else {
+                    current_jump += 14;
```
Mark-Simulacrum: I am not sure about these `+=` here but if it works in practice seems fine.

It might be nice to extract a test here and a function that given a date gives an iterator of dates (or maybe just a vector with, say, 20 dates).

src/main.rs
spastorino: But we don't :joy:, it's not really important given the order of magnitudes of things executing here. We could use `lazy_static` if we were going to do so.

src/main.rs
```diff
@@ -823,6 +865,66 @@ fn bisect(cfg: &Config, client: &Client) -> Result<(), Error> {
...
+    assert_eq!(start_date - chrono::Duration::days(8), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(15), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(22), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(29), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(36), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(43), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(50), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(64), iter.next().unwrap());
+    assert_eq!(start_date - chrono::Duration::days(78), iter.next().unwrap());
+}
```
Mark-Simulacrum: Amazing, thank you!

Mark-Simulacrum:
Feel free to merge once CI passes (I'm not sure what it tests beyond just compilation, but it exists, so we might as well go by it)

src/main.rs
spastorino: :+1:

Issue #34, opened 2019-07-14
Flag to check for internal compiler errors


fenhl:
It would be nice as a convenience if this could provide an `--ice` flag to check specifically for internal compiler errors rather than any error.

pnkfelix:
We definitely need this.

Specifically, such a flag should treat a compiler error as a "pass" and a ICE as a fail.

(we might need a separate mode that treats a compiler error as a "pass" and any other behavior (compiler acceptance or an ICE) as a fail. I'm not sure whether such work needs to be done up front, nor whether it would make sense to try to fold it into this task.)

chrissimpkins:
#47 adds a shell script that can be used to automate this until the functionality is implemented in the tool

pnkfelix:
Hmm. I've been working on putting support into `cargo-bisect-rustc` itself for this and a few other testing variations.

I've been assuming that building the support into `cargo-bisect-rustc` would be best for the long term, but maybe that is an incorrect assumption.

pnkfelix:
(the main reason I haven't put up a PR yet is due to the issue described by PR #45 )

chrissimpkins:
> Hmm. I've been working on putting support into cargo-bisect-rustc itself for this and a few other testing variations.
>
>I've been assuming that building the support into cargo-bisect-rustc would be best for the long term, but maybe that is an incorrect assumption.

Definitely no intent to derail that effort.  These were simple one-liner shell scripts and I figured that they might be of use to others until something is implemented in the tool.  I used the ICE stderr substring script in a few cargo bisect reports over the weekend - as an aside very much under the assumption that this is helpful / desirable from a new member of the "ICE-breaker Cleanup Crew" team :) - and it works well for now.

I spoke with Santiago about builtin subcommand functionality on Zulip today.  If that is something that you intend to do, I am definitely interested to lend a hand if there is anything that I can help with the effort.  I am relatively new to Rust and it might require a bit of mentorship at this stage.  But I have time and an interest if that helps to push it forward.

Issue #28, opened 2019-01-11
feature request: Invert status


nagisa:
Add ability to look for any change in success status. i.e. not only went from good to bad, but also from bad to good.

Mark-Simulacrum:
If I understand you correctly, I think this would probably require something more interesting than a binary search - we need the status to be monotonic (i.e. change only once). Perhaps I'm misunderstanding you though? We certainly can naively see if the status has changed under us and warn/print that...

nagisa:
Status changing only once is fine, its just that currently cargo-bisect-rustc requires the start end to always succeed and the end end to always fail.

nagisa:
Basically my goal today was to see where a bug was fixed, rather than appeared :)

Mark-Simulacrum:
I think I'd be happy to accept a PR for this change.. not sure I'll get a chance myself in the next few days.

spastorino:
@nagisa I know this is a bit old, trying to get this repo in shape a but. Is @kennytm suggestion enough? or are you up to provide a PR with your idea?

nagisa:
Using a script worked for me and I used the trick more than once since. I think this is a common enough want to implement it in bisect-rustc but I also don't think it is super critical.

PR #47, opened 2020-02-11
Add shell scripts for use with cargo-bisect-rustc

chrissimpkins:
This PR adds three shell scripts in a new `scripts` directory.  They support the following during bisections with this tool:

- `deicer.sh` - automate ICE identification during bisection with `cargo build`, returns exit status code 1 when the stderr substring `"error: internal compiler error"` is identified and ignores all other non-zero exit status code build fails (script workaround for https://github.com/rust-lang/cargo-bisect-rustc/issues/34). "regression" = ICE (only)
- `find.sh` - stdout/err stream substring search, returns exit status code 1 when stdout/err substring is identified.  "regression" = identified substring in stdout/err stream.
- `findfix.sh` - reverses the logic of the cargo build exit status to locate fixes - when build exits with 1, script exits with 0 and when build exits with 0, script exits with 1 (script workaround for https://github.com/rust-lang/cargo-bisect-rustc/issues/28 as recommended by [kennytm](https://github.com/rust-lang/cargo-bisect-rustc/issues/28#issuecomment-453554419)). "regression" = fix

If these are appropriate upstream, I'd be happy to add documentation in the tutorial or README.

### DIFF
--- a/scripts/find.sh
+++ b/scripts/find.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+# Copyright 2020 The Rust Project Developers
+#
+# Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+# http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+# http://opensource.org/licenses/MIT>, at your option. This file may not be
+# copied, modified, or distributed except according to those terms.
+
+NEEDLE="E0642"
+
+! cargo build 2>&1 | grep "$NEEDLE"
--- a/scripts/deicer.sh
+++ b/scripts/deicer.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+# Copyright 2020 The Rust Project Developers
+#
+# Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+# http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+# http://opensource.org/licenses/MIT>, at your option. This file may not be
+# copied, modified, or distributed except according to those terms.
+
+! cargo build 2>&1 | grep "error: internal compiler error"
--- a/scripts/findfix.sh
+++ b/scripts/findfix.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+# Copyright 2020 The Rust Project Developers
+#
+# Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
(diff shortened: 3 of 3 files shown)


scripts/find.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright 2020 The Rust Project Developers
+#
+# Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+# http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+# http://opensource.org/licenses/MIT>, at your option. This file may not be
+# copied, modified, or distributed except according to those terms.
+
+NEEDLE="E0642"
+
+! cargo build 2>&1 | grep "$NEEDLE"
```
hellow554: I would suggest to change this to use [parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) and let the variable set from the outer world, so this script would be more generic and you can do stuff like:

`env NEEDLE="E0600" scripts/find.sh`

instead of manually changing the script.

So something like:

```
! cargo build 2>&1 | grep "${NEEDLE:-E0642}"
```

should do the trick

chrissimpkins: Great idea! Thoughts about using a regex for any compiler error code as the default? Is that a common enough task or is there a better generally useful default fallback when env var is not set?

chrissimpkins:
The discussion about this PR forked into a few different threads... :) 

https://rust-lang.zulipchat.com/#narrow/stream/217417-t-compiler.2Fcargo-bisect-rustc/topic/Scripts.20for.20use.20with.20cargo-bisect-rustc

https://github.com/rust-lang/cargo-bisect-rustc/issues/34#issuecomment-584760241

My understanding of the consensus is that the ideal approach is through implementation in the Rust source rather than through external scripts.  I think that we all agree on this point. 

It sounds like @pnkfelix [plans to tackle the changes in the tool](https://github.com/rust-lang/cargo-bisect-rustc/issues/34#issuecomment-584760241).  I'd be happy to pitch in on the effort if there is anything that I can do.

scripts/find.sh
chrissimpkins: Holding on edits until https://github.com/rust-lang/cargo-bisect-rustc/pull/47#issuecomment-584904257 is settled.

pnkfelix:
@chrissimpkins FYI I finally posted my PR for ICE support (and more) in PR #53

chrissimpkins:
@pnkfelix ty!

PR #55, opened 2020-02-21
Link to current nightly

### LINKED ISSUE #54: nightly for `rustup default`  breaks cargo-bisectc-rustc when start..end range overaps current nightly
This is an issue to track the problem described on PR #45

Demonstration:

Here is a file that does *not* expose any regression in `rustc` (or `cargo`):

src/main.rs:
```rust
fn main() {
    println!("Hello world");
}
```

Here is a transcript of my interactions in the shell (note the use of `-v -v` to observe the actual runs and their output):

```
% rustup default
nightly-x86_64-apple-darwin (default)
% rustc --version
rustc 1.43.0-nightly (e620d0f33 2020-02-18)
% cargo-bisect-rustc --start 2020-02-17 -v -v
checking nightly-2020-02-17
std for x86_64-apple-darwin: 16.47 MB / 16.47 MB [=========] 100.00 % 4.91 MB/s    Compiling bisect-area v0.1.0 (/Users/felixklock/Dev/Mozilla/bisect-area)
    Finished dev [unoptimized + debuginfo] target(s) in 1.03s
uninstalling nightly-2020-02-17
verifying the start of the range does not reproduce the regression
std for x86_64-apple-darwin: 16.47 MB / 16.47 MB [=========] 100.00 % 4.57 MB/s    Compiling bisect-area v0.1.0 (/Users/felixklock/Dev/Mozilla/bisect-area)
    Finished dev [unoptimized + debuginfo] target(s) in 0.97s
uninstalling nightly-2020-02-17
tested nightly-2020-02-17, got No
confirmed the start of the range does not reproduce the regression
verifying the end of the range reproduces the regression
error: toolchain 'bisector-nightly-2020-02-18-x86_64-apple-darwin' is not installed
tested nightly-2020-02-18, got Yes
confirmed the end of the range reproduces the regression
searched toolchains nightly-2020-02-17 through nightly-2020-02-18
error: toolchain 'bisector-nightly-2020-02-18-x86_64-apple-darwin' is not installed
regression in nightly-2020-02-18
fetching https://static.rust-lang.org/dist/2020-02-18/channel-rust-nightly-git-commit-hash.txt
nightly manifest 2020-02-18: 40 B / 40 B [================] 100.00 % 66.08 KB/s converted 2020-02-18 to 0176a9eef845e7421b7e2f7ef015333a41a7c027
fetching https://static.rust-lang.org/dist/2020-02-17/channel-rust-nightly-git-commit-hash.txt
nightly manifest 2020-02-17: 40 B / 40 B [===============] 100.00 % 105.10 KB/s converted 2020-02-17 to 5e7af4669f80e5f682141f050193ab679afdb4b1
looking for regression commit between 2020-02-18 and 2020-02-17
```

This is the problematic part of the transcript:

```
verifying the end of the range reproduces the regression
error: toolchain 'bisector-nightly-2020-02-18-x86_64-apple-darwin' is not installed
tested nightly-2020-02-18, got Yes
```

That is, a internal bug in `cargo-bisect-rustc` leads to a problem with a particular `rustc +{} ...` invocation (where it substitutes `bisector-nightly-2020-02-18` for `{}`) and then `cargo-bisect-rustc` treats the failure of *that* invocation as a regression, rather than a problem external to `rustc`/`cargo`.

pnkfelix:
Fix code that was trying reuse an already installed nightly, by making `install` call out to `rustup toolchain link` to link the currently installed nightly up to the new "bisector-nightly-xxx" that we expect.

Fix #54

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -495,6 +495,14 @@ impl Toolchain {
 
     fn remove(&self, dl_params: &DownloadParams) -> Result<(), Error> {
         eprintln!("uninstalling {}", self);
+        self.do_remove(dl_params)
+    }
+
+    /// Removes the (previously installed) bisector rustc described by `dl_params`.
+    ///
+    /// The main reason to call this (instead of `fs::remove_dir_all` directly)
+    /// is to guard against deleting state not managed by `cargo-bisect-rustc`.
+    fn do_remove(&self, dl_params: &DownloadParams) -> Result<(), Error> {
         let rustup_name = self.rustup_name();
 
         // Guard aginst destroying directories that this tool didn't create.
@@ -711,7 +719,7 @@ impl Toolchain {
             .map_err(InstallError::TempDir)?;
         let dest = dl_params.install_dir.join(self.rustup_name());
         if dl_params.force_install {
-            let _ = fs::remove_dir_all(&dest);
+            let _ = self.do_remove(dl_params);
         }
 
         if dest.is_dir() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -505,7 +505,7 @@ impl Toolchain {
     fn do_remove(&self, dl_params: &DownloadParams) -> Result<(), Error> {
         let rustup_name = self.rustup_name();
 
-        // Guard aginst destroying directories that this tool didn't create.
+        // Guard against destroying directories that this tool didn't create.
         assert!(rustup_name.starts_with("bisector-nightly") ||
                 rustup_name.starts_with("bisector-ci"));
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,6 +31,7 @@ use std::env;
 use std::ffi::OsString;
 use std::fmt;
 use std::fs;
+use std::iter;
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use std::process::{self, Command, Stdio};
@@ -309,19 +310,16 @@ impl DownloadParams {
             if cfg.args.alt { "-alt" } else { "" }
         );
 
-        DownloadParams {
-            url_prefix: url_prefix,
-            tmp_dir: cfg.rustup_tmp_path.clone(),
-            install_dir: cfg.toolchains_path.clone(),
-            install_cargo: cfg.args.with_cargo,
-            install_src: cfg.args.with_src,
-            force_install: cfg.args.force_install,
-        }
+        Self::from_cfg_with_url_prefix(cfg, url_prefix)
     }
 
     fn for_nightly(cfg: &Config) -> Self {
+        Self::from_cfg_with_url_prefix(cfg, NIGHTLY_SERVER.to_string())
+    }
+
+    fn from_cfg_with_url_prefix(cfg: &Config, url_prefix: String) -> Self {
         DownloadParams {
-            url_prefix: NIGHTLY_SERVER.to_string(),
+            url_prefix: url_prefix,
             tmp_dir: cfg.rustup_tmp_path.clone(),
             install_dir: cfg.toolchains_path.clone(),
             install_cargo: cfg.args.with_cargo,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
This looks slightly wrong to me. The default toolchain could be `nightly-2020-01-04`, in which case our synthesis of the link path is wrong i believe. I suspect we should instead run rustc --print sysroot or so and get the path from there.

I was also concerned about removing someone's installed nightly, but it looks like remove_dir_all isn't supposed to follow symlinks (though I'm not sure what behavior is like for hardlinks, I believe rustup link doesn't create those). I guess modulo https://github.com/rust-lang/rust/issues/48504 but hopefully we don't hit that.

pnkfelix:
> The default toolchain could be `nightly-2020-01-04`, in which case our synthesis of the link path is wrong i believe. I suspect we should instead run rustc --print sysroot or so and get the path from there.

Yeah I just realized this while I was experimenting with this and happened to be making local changes to my `rustup`'s default nightly. Let me see if I can fix it.

(It seems like `rustup` is missing some functionality here, no? I.e. I could imagine we "need" a subcommand like `rustup toolchain print-install-dir <toolchain>` or something along those lines...)

Mark-Simulacrum:
Yes, rustup isn't really ready for scripting today, semi-intentionally. Or so I recall hearing from @Kinnison... I think.

src/main.rs
```diff
@@ -494,11 +494,23 @@ impl Toolchain {
...
+    }
+
+    /// Removes the (previously installed) bisector rustc described by `dl_params`.
+    ///
+    /// The main reason to call this (instead of `fs::remove_dir_all` directly)
+    /// is to guard against deleting state not managed by `cargo-bisect-rustc`.
+    fn do_remove(&self, dl_params: &DownloadParams) -> Result<(), Error> {
+        let rustup_name = self.rustup_name();
+
+        // Guard aginst destroying directories that this tool didn't create.
```
spastorino:
```suggestion
        // Guard against destroying directories that this tool didn't create.
```

src/main.rs
```diff
@@ -494,11 +494,23 @@ impl Toolchain {
...
+        assert!(rustup_name.starts_with("bisector-nightly") ||
+                rustup_name.starts_with("bisector-ci"));
```
spastorino: I understand this idea and seems good but given that we are naming the default installation as nightly https://github.com/rust-lang/cargo-bisect-rustc/pull/55/files#diff-639fbc4ef05b315af92b4d836c31b023R737 I don't see how are we avoiding to call do_remove for the default nightly installations.

pnkfelix: (hmm unfortunately I just killed that hyperlink via my force push ...)

pnkfelix:
(just forced-push a new version that uses `rustc --print sysroot` to figure out the directory to link to, instead of trying to (incorrectly) infer it.)

src/main.rs
pnkfelix: @spastorino can you provide an updated link with the new code I force pushed before seeing your comment? I'm not sure what point you are making and I think having the context will help.

(In particular, when you say "the default installation", I do not know if you are referring to `rustup`'s default installation or to ... `cargo-bisect-rustc's` ?

In any case, `cargo-bisect-rustc` certainly shouldn't be calling `do_remove` for any nightly installations. If we are doing that currently, it sounds like a bug.

pnkfelix:
> I was also concerned about removing someone's installed nightly, but it looks like remove_dir_all isn't supposed to follow symlinks

You know, a way to definitely avoid this risk is to "just" fully copy the directory tree locally, rather than use a symbolic link. And then the resulting directory structure for the bisector variants would be totally uniform, and it would also avoid the risk (when using `--preserve`) of a user independently removing an installed nightly that we linked to...

(it would also be wasteful of disk space, of course... but that might be a micro-optimization for something like this.)

Its at least something to consider, though I would be happy just landing this PR as is, rather than bikeshedding this question.

src/main.rs
spastorino: My comment doesn't seem to apply anymore. Let me review again with the updated code.

PR #56, opened 2020-02-22
Edit help strings

chrissimpkins:
Establishes consistent help menu casing and makes some strings more concise to fit on a single line in 80 column views

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -90,10 +90,9 @@ struct Opts {
     #[structopt(
         long = "regress",
         default_value = "error",
-        help = "Customize what is treated as regression.",
-        long_help = "Customize what is treated as regression. \
-                     Values include `--regress=error`, `--regress=non-error`, \
-                     `--regress=ice`, and `--regress=success`."
+        help = "Custom regression definition",
+        long_help = "Custom regression definition \
+                     [error|non-error|ice|success]"
     )]
     regress: String,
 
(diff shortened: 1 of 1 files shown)


pnkfelix:
Can you post a similar rendering of what it looks like with the long help output? I just want to double-check that things like `--regress` are still relatively clear in their usage.

chrissimpkins:
#### Before



#### After (80 columns)



#### After (90 columns)

chrissimpkins:
I tried to fix the newline issue with your long help string but it seems to be something on the clap side (I think?).  I couldn't get it to format short source lines with escaped newlines AND eliminate the newline in the help output.

pnkfelix:
@chrissimpkins I'm glad I'm not the only one who felt stymied by the weirdness of the newline issue. (Or at least, by "newline issue", I assume you mean the fact that the old rendered output fails to include a blank line between the entries for `--regress` and `--script`.

Anyway, looks good to me.

chrissimpkins:
Different issue.  There is a newline in the long help string between the `--regress=non-error` and `--regress=ice` (see before image in https://github.com/rust-lang/cargo-bisect-rustc/pull/56#issuecomment-590356107).  IIRC that occurred at an escaped newline in the StructOpt source for the longhelp definition.

chrissimpkins:
And thanks for merging this!

PR #57, opened 2020-02-22
Transition to Edition 2018

chrissimpkins:
This PR transitions the project to edition 2018

Includes:

- modify Cargo.toml with `edition = "2018"` field
- remove unnecessary extern crate declarations
- add explicit `crate` root path on use statement
- add explicit anonymous lifetime idioms with `cargo fix --edition-idioms`
- transition `#[macro_use]` attributes to use statements


pnkfelix:
Thanks @chrissimpkins !

PR #60, opened 2020-02-26
Dependency updates

chrissimpkins:
Updates dependencies to current release versions.

Changes that required updates:

- `reqwest` now defaults to async GET requests that do not support the Read trait (required here).  The blocking GET request approach is now supported in the `reqwest::blocking` library.  The use statement was updated in `main.rs`
- `structopt` no longer uses the `raw` attribute.  This was reformatted in https://github.com/rust-lang/cargo-bisect-rustc/commit/b0790373730c5129343de629e473318324ad256b

Notable dependency changes:

-  It looks like`tempdir` is a [deprecated library](https://crates.io/crates/tempdir). The source is now in the `tempfile` library.  There is a slightly different API for the creation of a named temp directory within another directory.  It looks like this will require use of the [`Builder` struct](https://docs.rs/tempfile/3.1.0/tempfile/struct.Builder.html):

```rust
let tmpdir = Builder::new().prefix(SUBDIR_PATH).tempdir_in(TEMPDIR_PATH)?;
```

I updated to the most recent `tempdir` crate release and did not transition to `tempfile`.  Happy to do so if you want to move to the new crate.


spastorino:
Seems good :). Please update the lock file too.

chrissimpkins:
> Please update the lock file too.

ty!

https://github.com/rust-lang/cargo-bisect-rustc/pull/60/commits/22a0c3e821dec3628cf649b8a266a964cc01b7d1

Issue #40, opened 2019-12-06
Using the Github API instead of a local git repository


hellow554:
If you want to use this tool you either specify a local git repository **at compile time** or it will always clone a bare repository on every `cargo-bisect-rustc` run. 

This is somewhat inconvenient and using the Github API is something one could do instead.

What are your thoughts about it? I'd like to implement it, maybe as a flag (`--no-local-git-repo-is-available-so-use-github-instead-plz`)

Mark-Simulacrum:
If we can reliably use the GitHub API (without, e.g., hitting rate limits) then I would love to do that by default (with perhaps a flag to fallback to local git repo). Happy to accept a PR!

spastorino:
:+1: to this idea too.

pnkfelix:
I have a [prototype](https://github.com/pnkfelix/cargo-bisect-rustc/tree/use-github-api) of this that I need to rebase.

PR #62, opened 2020-02-27
Refactor Toolchain testing, eliminate duplicated start date checks

### LINKED ISSUE #61: Nightly testing start date test is repeated when --start is defined on command line
Test source with regression from [blog post](https://blog.rust-lang.org/inside-rust/2019/12/18/bisecting-rust-compiler.html):

```rust
pub struct Slice<'a, T>(&'a [T]);

impl<'a, T: 'a> Slice<'a, T> {
    pub const EMPTY: Self = Slice ({
        let v: &[T] = &[];
        v
    });
}

fn main() {
    let s = Slice(&[1, 2]);
    assert!(s.0 != Slice::EMPTY.0);
}
```

Test:

```
$ cargo bisect-rustc --start=2019-09-20 --end=2019-10-02

checking nightly-2019-09-20
std for x86_64-apple-darwin: 172.68 MB / 172.68 MB [==========================] 100.00 % 18.01 MB/s uninstalling nightly-2019-09-20
verifying the start of the range does not reproduce the regression
std for x86_64-apple-darwin: 172.68 MB / 172.68 MB [==========================] 100.00 % 18.07 MB/s uninstalling nightly-2019-09-20
tested nightly-2019-09-20, got No
confirmed the start of the range does not reproduce the regression
verifying the end of the range reproduces the regression
std for x86_64-apple-darwin: 171.05 MB / 171.05 MB [==========================] 100.00 % 17.30 MB/s uninstalling nightly-2019-10-02
tested nightly-2019-10-02, got Yes
confirmed the end of the range reproduces the regression

[ ... snip ... ]
```

The 2020-09-20 nightly is installed, tested, & uninstalled twice.  The first test is in the `bisect_nightlies` function.  The second test is in the `least_satisfying` function.

It looks like the duplication happens because the bisect commit testing does rely on the start range test in `least_satisfying`.

chrissimpkins:
Fixes #61 

Test source with regression:

```rust
pub struct Slice<'a, T>(&'a [T]);

impl<'a, T: 'a> Slice<'a, T> {
    pub const EMPTY: Self = Slice ({
        let v: &[T] = &[];
        v
    });
}

fn main() {
    let s = Slice(&[1, 2]);
    assert!(s.0 != Slice::EMPTY.0);
}
```

This PR eliminates the duplication of the nightly start date check that is described in #61.

---

### TODO

- [x] refactor start date validation for absence of regression to new function (https://github.com/rust-lang/cargo-bisect-rustc/pull/62#issuecomment-592138451)
- [x] +/- refactor end date validation for presence of regression to new function


spastorino:
What if we just extract the verify start code as a separate function?.

chrissimpkins:
> What if we just extract the verify start code as a separate function?.

SGTM.   

Thoughts about doing the same with the end of range validation that confirms that the regression is present?  This would move the validations out of `least_satisfying`.  I think that we can then call the validation functions from [`bisect_nightlies`](https://github.com/rust-lang/cargo-bisect-rustc/blob/1ed199da35d32fc1698eff3673868b5a5829071f/src/main.rs#L1215) and [`bisect_ci_between`](https://github.com/rust-lang/cargo-bisect-rustc/blob/1ed199da35d32fc1698eff3673868b5a5829071f/src/main.rs#L1380)?

chrissimpkins:
I intend to refactor this but will wait for @pnkfelix to merge [the Github API support](https://github.com/rust-lang/cargo-bisect-rustc/issues/40#issuecomment-592851637).  I think that the changes that I make will lead to a more significant rebase of that work.   Will return to this.

chrissimpkins:
rebased on 25e566a3eb5645b3100d9e07cfe88a1f184c5074

chrissimpkins:
Working on https://github.com/rust-lang/cargo-bisect-rustc/pull/62#issuecomment-592148668

chrissimpkins:
As of https://github.com/rust-lang/cargo-bisect-rustc/pull/62/commits/9c49536f6b5f787abc0d6be2a9d4477a8ed24140 , I've included a fairly extensive refactor of the:

- start bound validation checks
- end bound validation checks
- stderr reporting during execution

This includes deduplication of the `least_satisfying` code in the nightly and CI bisection routines and stderr reporting source across the installs, uninstalls, and testing.

The nightly stderr reporting is now formatted like this:

```
installing nightly-2019-09-15
std for x86_64-apple-darwin: 173.39 MB / 173.39 MB [=======================] 100.00 % 16.98 MB/s 0s testing...
RESULT: nightly-2019-09-15, ===> No
uninstalling nightly-2019-09-15

installing nightly-2019-10-02
std for x86_64-apple-darwin: 171.05 MB / 171.05 MB [=======================] 100.00 % 17.40 MB/s 0s testing...
RESULT: nightly-2019-10-02, ===> Yes
uninstalling nightly-2019-10-02

installing nightly-2019-09-23
std for x86_64-apple-darwin: 172.64 MB / 172.64 MB [==========================] 100.00 % 18.00 MB/s testing...
RESULT: nightly-2019-09-23, ===> No
uninstalling nightly-2019-09-23

installing nightly-2019-09-27
std for x86_64-apple-darwin: 173.56 MB / 173.56 MB [==========================] 100.00 % 17.94 MB/s testing...
RESULT: nightly-2019-09-27, ===> No
uninstalling nightly-2019-09-27
```

and the CI stderr is formatted in a similar fashion:

```
installing 488381ce9ef0ceabe83b73127c659e5d38137df0
std for x86_64-apple-darwin: 170.66 MB / 170.66 MB [===========================] 100.00 % 9.83 MB/s testing...
RESULT: 488381ce9ef0ceabe83b73127c659e5d38137df0, ===> No
uninstalling 488381ce9ef0ceabe83b73127c659e5d38137df0

installing 8431f261dd160021b6af85916f161a13dd101ca0
std for x86_64-apple-darwin: 171.03 MB / 171.03 MB [===========================] 100.00 % 3.77 MB/s testing...
RESULT: 8431f261dd160021b6af85916f161a13dd101ca0, ===> Yes
uninstalling 8431f261dd160021b6af85916f161a13dd101ca0

installing fe2f7e0e5350a431e00e72cfd00df2669f94fe06
std for x86_64-apple-darwin: 170.60 MB / 170.60 MB [===========================] 100.00 % 4.12 MB/s testing...
RESULT: fe2f7e0e5350a431e00e72cfd00df2669f94fe06, ===> Yes
uninstalling fe2f7e0e5350a431e00e72cfd00df2669f94fe06

installing b61e69433951e31f7bd746e22f516a48ad41623b
std for x86_64-apple-darwin: 170.53 MB / 170.53 MB [===========================] 100.00 % 3.29 MB/s testing...
RESULT: b61e69433951e31f7bd746e22f516a48ad41623b, ===> No
uninstalling b61e69433951e31f7bd746e22f516a48ad41623b

installing 0bbab7d99dde8620604fb265706dc8bff20345a7
std for x86_64-apple-darwin: 170.60 MB / 170.60 MB [===========================] 100.00 % 6.99 MB/s testing...
RESULT: 0bbab7d99dde8620604fb265706dc8bff20345a7, ===> Yes
uninstalling 0bbab7d99dde8620604fb265706dc8bff20345a7

searched toolchains 488381ce9ef0ceabe83b73127c659e5d38137df0 through 8431f261dd160021b6af85916f161a13dd101ca0
regression in 0bbab7d99dde8620604fb265706dc8bff20345a7
```

The duplicated start (nightly) date checks are eliminated as of https://github.com/rust-lang/cargo-bisect-rustc/pull/62/commits/9c49536f6b5f787abc0d6be2a9d4477a8ed24140

chrissimpkins:
> What if we just extract the verify start code as a separate function?.

@spastorino I ended up refactoring the validation and bisection code into two functions (one for individual toolchain testing, one for the call to the `least_satisfying` function for the bisection runs) that are reused throughout the nightly and CI testing.

src/least_satisfying.rs
```diff
@@ -8,7 +8,7 @@
 use std::collections::BTreeMap;
 use std::fmt;
 
-pub fn least_satisfying<T, P>(slice: &[T], mut predicate: P) -> usize
+pub fn least_satisfying<T, P>(slice: &[T], start_check: bool, mut predicate: P) -> usize
```
pnkfelix: can you add a comment here stating what the setting of `start_check` entails?

Something like this would suffice (asuming it is an accurate description):
> if `start_check` is false, then the caller has already ensured that  the start of the range does not reproduce the regression (i.e. that running the `predicate` at the start date returns `Satisfies::No`). If `start_check` is true, then this method will itself confirm that running  the `predicate` at the start date returns `Satisfies::No`.

pnkfelix: Oh,  I now see that this parameter may go away in a later commit.

Assuming the `start_check` parameter does go away in a later commit, then there is no need to document it here.

chrissimpkins: That new parameter that was in early commits in this PR should be gone after the refactor beginning at ~https://github.com/rust-lang/cargo-bisect-rustc/pull/62/commits/e50052ec966ac6d58c28d02dd1572690ef751b11

I was initially using it to prevent the duplicated nightly start date checks but I ended up refactoring the entire approach based on Santiago's suggestion.

src/least_satisfying.rs
```diff
@@ -105,111 +93,111 @@ mod tests {
...
     fn least_satisfying_1() {
         assert_eq!(
-            least_satisfying(&[No, Unknown, Unknown, No, Yes], true,|i| *i),
+            least_satisfying(&[No, Unknown, Unknown, No, Yes],|i| *i),
             4
         );
     }
 
     #[test]
     fn least_satisfying_1f() {
```
pnkfelix: these  variant tests you added in the previous commit, which as far as I can tell were to ensure that the introduction of `start_check` did not misbehave, now serve no purpose. IMO you can eliminate them, either from this commit, or from the earlier commit that introduced `start_check`.

src/main.rs
```diff
@@ -792,51 +844,61 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> Result<BisectionResult, Er
...
+            Ok(r) => {
+                // If Satisfies::No, then the regression was not identified in this nightly.
+                // Break out of the loop and use this as the start date for the
+                // bisection range
+                if let Satisfies::No = r {
                     first_success = Some(nightly_date);
                     break;
                 } else if has_start {
-                    bail!("the start of the range to test must not reproduce the regression");
+                    // If this date was explicitly defined on CL & has regression,
```
pnkfelix: I don't think the abbreviation "CL" is common enough to use it here. Please spell out "command line", assuming that *is* what this stood for.

pnkfelix: (you might win me over if you wrote "CLI" instead of  "CL"; but the best answer is to not use an acronym at all here.)

src/least_satisfying.rs
chrissimpkins: They will go away as you work your way through the commits... :)

src/main.rs
chrissimpkins: Will do!

src/least_satisfying.rs
pnkfelix: nope, they are in the diff. :)

src/main.rs
chrissimpkins: No problem to write out "command line".  I will do that

src/least_satisfying.rs
chrissimpkins: Hmmm... OK let me check.  Apologies.

pnkfelix:
Okay, this looks fine.

r=me once you've addressed my nits about comments and gotten rid of the tests that had been added solely to check the `start_check` parameter which subsequently went away.

src/least_satisfying.rs
pnkfelix: (maybe you have another local commit that you still need to push up to github?)

chrissimpkins: I am showing that I removed it in 8227356a2bf0640fc1eeb427dea568d544ead77d

src/least_satisfying.rs
```diff
@@ -105,111 +93,111 @@ mod tests {
...
+        let res = least_satisfying(&satisfies_v, |i| *i);
         let exp = first_yes.unwrap();
         TestResult::from_bool(res == exp)
     }
 
     #[test]
     fn least_satisfying_1() {
         assert_eq!(
-            least_satisfying(&[No, Unknown, Unknown, No, Yes], true,|i| *i),
+            least_satisfying(&[No, Unknown, Unknown, No, Yes],|i| *i),
```
chrissimpkins: @pnkfelix here

chrissimpkins: Let me confirm that I didn't add it back for some reason

chrissimpkins:
Thank you Felix!  I will have a detailed look through it tonight and update to address your review.  I really appreciate it.

src/least_satisfying.rs
chrissimpkins: Sorry, sorry.  I misunderstood this issue.  Yes, I do need to remove the duplicated tests.  I thought that you were pointing to the boolean that was added to the tests.  I forgot that I duplicated all of them.  I will definitely remove them.

chrissimpkins: d1e7eb8fce8afb9f713a85b372462e4aab6f8082

src/main.rs
chrissimpkins: c950c79319382cb9beff5dacace48ca3ebc23d06

chrissimpkins:
r? @pnkfelix 

Pushed changes to address your review in https://github.com/rust-lang/cargo-bisect-rustc/pull/62/commits/d1e7eb8fce8afb9f713a85b372462e4aab6f8082 & https://github.com/rust-lang/cargo-bisect-rustc/pull/62/commits/c950c79319382cb9beff5dacace48ca3ebc23d06

PR #63, opened 2020-03-02
Add support for Github API for querying repository info

### LINKED ISSUE #40: Using the Github API instead of a local git repository

pnkfelix:
Adds `--access` command line flag.

There are two values current for the flag: `--access=checkout` (the original behavior, and remains the default for now) and `--access=github (the new behavior).

In the long term, we will probably want `--access=github` to be the default, but there are some refinements I think we want before we do that.

In any case, I wanted to get this code up so that it doesn't bitrot further, since it seems like there are parallel efforts that could accelerate bitrot if I don't get it up.

----

The aforementioned refinements are written out in one of the commits, but here they are again:

 1. To-do: Fallback to local checkout if the github accessor errors out

 2. To-do: Add a warning to stdout if the user runs github access without  passing an access token via GITHUB_TOKEN env var (since doing that will    hit rate limit much faster).

----

Fix #40

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -1,0 +1,167 @@
+use failure::Error;
+use reqwest::{self, blocking::Client, blocking::Response};
+use serde::{Deserialize, Serialize};
+
+use crate::Commit;
+
+#[derive(Serialize, Deserialize, Debug)]
+struct GithubCommitElem { commit: GithubCommit, sha: String }
+#[derive(Serialize, Deserialize, Debug)]
+struct GithubCommit { author: GithubAuthor, committer: GithubAuthor, message: String, }
+#[derive(Serialize, Deserialize, Debug)]
+struct GithubAuthor { date: String, email: String, name: String }
+
+type GitDate = chrono::DateTime<chrono::Utc>;
+
+impl GithubCommitElem {
+    fn date(&self) -> Result<GitDate, Error> {
+        Ok(self.commit.committer.date.parse()?)
+    }
+
+    fn git_commit(self) -> Result<Commit, Error> {
+        let date = self.date()?;
+        Ok(Commit {
+            sha: self.sha,
+            date,
+            summary: self.commit.message,
+        })
+    }
+}
+
+fn headers() -> Result<reqwest::header::HeaderMap, Error> {
+    let mut headers = reqwest::header::HeaderMap::new();
+    let user_agent = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+    let user_agent = reqwest::header::HeaderValue::from_static(user_agent);
+    headers.insert(reqwest::header::USER_AGENT, user_agent);
+    if let Ok(token) = std::env::var("GITHUB_TOKEN") {
+        eprintln!("adding local env GITHUB_TOKEN value to headers in github query");
+        let value = reqwest::header::HeaderValue::from_str(&format!("token {}", token))?;
+        headers.insert(reqwest::header::AUTHORIZATION, value);
+    }
+    Ok(headers)
+}
+
+pub(crate) fn get_commit(sha: &str) -> Result<Commit, Error> {
+    let url = SingleCommitUrl { sha }.url();
+    let client = Client::builder()
+        .default_headers(headers()?)
+        .build()?;
+    let response: Response = client.get(&url).send()?;
+    let elem: GithubCommitElem = response.json()?;
+    elem.git_commit()
+}
+
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct CommitsQuery<'a> {
+    pub since_date: &'a str,
+    pub most_recent_sha: &'a str,
+    pub earliest_sha: &'a str,
+}
+
+/// Returns the bors merge commits between the two specified boundaries
+/// (boundaries inclusive).
+
+impl<'a> CommitsQuery<'a> {
+    pub fn get_commits(&self) -> Result<Vec<Commit>, Error> {
+        get_commits(*self)
+    }
+}
+
+const PER_PAGE: usize = 100;
+const OWNER: &'static str = "rust-lang";
+const REPO: &'static str = "rust";
+
+
+trait ToUrl { fn url(&self) -> String; }
+struct CommitsUrl<'a> { page: usize, author: &'a str, since: &'a str, sha: &'a str }
+struct SingleCommitUrl<'a> { sha: &'a str }
+
+impl<'a> ToUrl for CommitsUrl<'a> {
+    fn url(&self) -> String {
+        format!("https://api.github.com/repos/{OWNER}/{REPO}/commits\
+                 ?page={PAGE}&per_page={PER_PAGE}\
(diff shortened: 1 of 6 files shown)


src/github.rs
```diff
@@ -0,0 +1,167 @@
...
+        })
+    }
+}
+
+fn headers() -> Result<reqwest::header::HeaderMap, Error> {
+    let mut headers = reqwest::header::HeaderMap::new();
+    let user_agent = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+    let user_agent = reqwest::header::HeaderValue::from_static(user_agent);
+    headers.insert(reqwest::header::USER_AGENT, user_agent);
+    if let Ok(token) = std::env::var("GITHUB_TOKEN") {
```
chrissimpkins: So long as the reqwest GET request specifies a `https://` URL, are the headers are always sent securely?  This is a naive question.  I don't know the reqwest library or any of its TLS dependencies well.  It looks like it uses multiple platform-specific TLS backends by default.

src/main.rs
```diff
@@ -31,10 +31,18 @@ use tee::TeeReader;
...
 use xz2::read::XzDecoder;
 
-mod git;
 mod least_satisfying;
+mod repo_access;
 
 use crate::least_satisfying::{least_satisfying, Satisfies};
+use crate::repo_access::{AccessViaGithub, AccessViaLocalGit, RustRepositoryAccessor};
+
+#[derive(Debug, Clone, PartialEq)]
```
chrissimpkins: Thoughts about impl Display on the Commit struct?

src/main.rs
```diff
@@ -159,32 +162,36 @@ struct Opts {
     script: Option<PathBuf>,
 }
 
+pub type GitDate = Date<Utc>;
+
 #[derive(Clone, Debug)]
 enum Bound {
     Commit(String),
```
chrissimpkins: And maintaining a connection between the `Commit` struct and the `Bound::Commit` enum or a link back to the GH data pulled into the `Commit` struct?  IMO it would be helpful to have the commit message string for the final report, particularly when this lands on a rollup PR commit where rolled up PR sha strings are included in the git commit message.

spastorino:
Great work @pnkfelix this looks very good.

Will wait for you to address @chrissimpkins comments and then I'd merge it :).

src/github.rs
pnkfelix: Honestly, I don't know the answer here either. It would be a good thing to double-check.

src/main.rs
pnkfelix: hmm.  My interpretation is that `Bound` captures the specification fed in by the user. I guess you're suggesting that we ... expand it, so that it would accumulate information as we acquire it from the commit messages in the repository?

(or maybe I misunderstand what kind of connection you are asking for here? Do you want the information stored inline, or some sort of hashmap?)

Anyway I do not think we need to block this PR on this particular suggested changes. We can explore this question after this lands.

src/github.rs
pnkfelix: (Resolving this *is* an example of something that is worth blocking this PR on, I think. I definitely want to be careful about leaking `GITHUB_TOKEN` over insecure channels...)

chrissimpkins: Who is the reqwest guru?

src/main.rs
chrissimpkins: > Anyway I do not think we need to block this PR on this particular suggested changes. We can explore this question after this lands.

Definitely agree. Just a comment now that you have additional data stored in a data structure from the GH API pulls.  It would be useful IMO for these data to move through the testing execution and be available at the final report generation stage.  I think that we lose the `Commit` struct early in the path and the `Bound` data structure is what is available for report generation.  My convoluted comment above was simply wondering out loud if there is a way to connect those pieces so that we have the commit messages for use in the report.  Maybe this is not necessary / is low priority?  GH shows them in popups when you hover over a commit SHA and we *do* have that in the report.

src/github.rs
Mark-Simulacrum: AFAIK, this should be fine. reqwest to my knowledge does the right thing by default.

One thing to be a bit cautious with is that the debug formatting of the request will include the token, but we're not publishing that I think so hopefully not a problem.

chrissimpkins: 👍

src/main.rs
pnkfelix: Okay, so lets attack this in a separate follow-up task.

src/github.rs
pnkfelix: > One thing to be a bit cautious with is that the debug formatting of the request will include the token

(Yeah, this is a classic example of where having a `fmt::Display` that delegates to `fmt::Debug` is an anti-pattern. Because it *is* reasonable to want to view the `GITHUB_TOKEN` info as part of local debugging; but you also don't to risk it leaking out into a separate `format!("{}", ...)` invocation...)

src/main.rs
pnkfelix: Filed as https://github.com/rust-lang/cargo-bisect-rustc/issues/64

src/github.rs
pnkfelix: It sounds like this is resolved then, yes?


----

Out of some degree of paranoia, I surveyed the source code for `reqwest` and didn't notice any obviously wrong uses of `{:?}`.

The two main places I. saw `{:?}` used in the `reqwest` source code for things other than strings and URIs were in:

 1. `#[test]` methods (which is obviously fine)
 2. `log::trace` calls (which is ... probably fine...?)

src/main.rs
pnkfelix: (lets table this to issue #64)

src/github.rs
Mark-Simulacrum: Yes, I think this is resolved. I didn't type this up earlier, but to be clear, this *can* go wrong, but realistically only if the user has misconfigured their system pretty badly -- e.g. adding some random certificates to their trusted list. And even then I'm not entirely sure the native-tls stack will pick it up by default; I think it should, since reqwest (or dependencies) don't ship a certificate bundle to my knowledge.

But that seems like a threat vector we do not care about; if you've done that you're in a not great place and likely leaking a github token as a result is the least of your worries :)

pnkfelix:
@spastorino okay I think everything has been resolved now.

(I am separately working on a regression test infrastructure; that side-project has given me sufficient confidence that this does seem to be working. I'll be posting the test system in a separate PR later; its taking me longer to polish than I expected.)

pnkfelix:
> Great work @pnkfelix this looks very good.
> 
> Will wait for you to address @chrissimpkins comments and then I'd merge it :).

Actually, I'm going to interpret this as an implicit "r=me" from @spastorino and go ahead and merge this myself.

PR #65, opened 2020-03-04
Add --regress=non-ice

Alexendoo:
Used to find when an ICE has been fixed, e.g. helpful for https://github.com/rust-lang/glacier

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,7 +73,7 @@ struct Opts {
         default_value = "error",
         help = "Custom regression definition",
         long_help = "Custom regression definition \
-                     [error|non-error|ice|success]"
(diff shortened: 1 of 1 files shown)


spastorino:
Unsure I like `non-ice` as a name for this. Anyway, will leave this up to @pnkfelix who has setup all this :)

pnkfelix:
There was some reason that I had originally thought this was unnecessary, but looking over the categories now, I don't know why I thought that.  Notably, `--regress=success` does not cover the case where you fix an ICE by replacing it with a static error; none of the existing cases do.

`--regress=non-ice` *will* cover that case. (It will also cover fixing an ICE by replacing it with a compilation success, which is also covered by `--regress=success`)

So, this seems fine to me.

PR #72, opened 2020-04-03
Format code using 'cargo fmt'

Atul9:

PR #73, opened 2020-04-04
Add cargo fmt to travis build config

Atul9:
To fix the build failure, https://github.com/rust-lang/cargo-bisect-rustc/pull/72 PR needs to be merged. I can rebase with master once https://github.com/rust-lang/cargo-bisect-rustc/pull/72 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,12 @@ rust:
   - nightly
 
 script:
+  - |
+    if [[ $TRAVIS_RUST_VERSION == *stable* ]]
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,5 +5,12 @@ rust:
   - nightly
 
 script:
```
chrissimpkins, line 7: Is it worth considering a build matrix that uses the toolchain rustfmt rather than stable only?

PR #75, opened 2020-04-11
Fix git commit retrieval via GitHub API when `--end` commit SHA is not specified

### LINKED ISSUE #74: Missing --end definition with SHA testing leads to error under --access=github approach
Related #71 

#71 fixes this bug when the Rust repository is cloned locally.  When the GitHub API is used with the `--access=github` argument, a `--start` definition is defined without an `--end` definition, and the request starts execution at the commit SHA level, the error is still present:

```
$ cargo bisect-rustc --start=488381ce9ef0ceabe83b73127c659e5d38137df0 --access=github
bisecting ci builds
starting at 488381ce9ef0ceabe83b73127c659e5d38137df0, ending at origin/master
adding local env GITHUB_TOKEN value to headers in github query
ERROR: error decoding response body: missing field `commit` at line 1 column 141
```

Test source from the [Inside Rust tutorial](https://blog.rust-lang.org/inside-rust/2019/12/18/bisecting-rust-compiler.html).

chrissimpkins:
Fixes #74 

Also updates project dependencies to current release versions and adds changelog updates for a v0.4.1 release.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -117,7 +117,7 @@ impl<'a> ToUrl for SingleCommitUrl<'a> {
         // HEAD commit as the end commit in this case
         if self.sha == "origin/master" {
             format!(
-                "https://api.github.com/repos/{OWNER}/{REPO}/commits/HEAD",
+                "https://api.github.com/repos/{OWNER}/{REPO}/commits/master",
                 OWNER = OWNER,
                 REPO = REPO,
             )
--- a/src/github.rs
+++ b/src/github.rs
@@ -112,12 +112,23 @@ impl<'a> ToUrl for CommitsUrl<'a> {
 
 impl<'a> ToUrl for SingleCommitUrl<'a> {
     fn url(&self) -> String {
-        format!(
-            "https://api.github.com/repos/{OWNER}/{REPO}/commits/{REF}",
-            OWNER = OWNER,
-            REPO = REPO,
-            REF = self.sha
-        )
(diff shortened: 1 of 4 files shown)


src/github.rs
```diff
@@ -112,12 +112,23 @@ impl<'a> ToUrl for CommitsUrl<'a> {
...
-            OWNER = OWNER,
-            REPO = REPO,
-            REF = self.sha
-        )
+        // "origin/master" is set as `sha` when there is no `--end=` definition
+        // specified on the command line.  We define the GitHub master branch
+        // HEAD commit as the end commit in this case
+        if self.sha == "origin/master" {
+            format!(
+                "https://api.github.com/repos/{OWNER}/{REPO}/commits/HEAD",
```
pnkfelix, line 120: would `commits/master` here work just as well?

if so, I think that would be clearer. (I'll be honest, I'm not clear on what `HEAD` is supposed to denote when querying `github`...)

chrissimpkins: > would commits/master here work just as well?

Yes it looks like it:

https://api.github.com/repos/rust-lang/rust/commits/master

https://api.github.com/repos/rust-lang/rust/commits/HEAD

chrissimpkins: Pushed this change in be54918d5e80f1cccc18de0e7885a1101d74f1c2

chrissimpkins:
Ty!

PR #78, opened 2020-04-20
Include compiler crashes in ICE defintion

Alexendoo:
This matches the definition [used in glacier], which includes exit status
101 without an accompanying diagnostic (such as rust-lang/rust#21599), and when
rustc is killed by a signal (rust-lang/rust#13368)

This also means no processing modes are capturing stdio, but I didn't
remove it as it may be desired for the regex features discussed in #53

[used in glacier]: https://github.com/rust-lang/glacier/blob/77029d8e7f755bd95913d3c741738674d5ccadc3/src/lib.rs#L51-L56

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -239,39 +239,33 @@ impl Config {
             status, stdout_utf8, stderr_utf8
         );
 
-        let saw_ice = || -> bool { stderr_utf8.contains("error: internal compiler error") };
+        // for commit message:
+        // no text but 101 https://github.com/rust-lang/rust/issues/21599
+        // no text, signal https://github.com/rust-lang/rust/issues/13368
 
-        let input = (self.output_processing_mode(), status.success());
+        const SUCCESS: Option<i32> = Some(0);
+        const ICE: Option<i32> = Some(101);
+
+        let input = (self.output_processing_mode(), status.code());
         let result = match input {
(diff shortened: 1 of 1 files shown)


pnkfelix:
I don't think this PR was quite right; it treats error code 101 as a unique identifier for ICE's, but even normal static errors from `rustc` will produce that same error code, I believe.

This mistake ends up breaking certain uses of `--regress=ice`, such as one I'm trying to include in my test suite as part of PR #88.

I'm going to revert this PR as part of PR #88 and open an issue to re-add its effects in a manner compatible with `--regress=ice`

pnkfelix:
I do see that Glacier *is* [using this same logic](https://github.com/rust-lang/glacier/blob/77029d8e7f755bd95913d3c741738674d5ccadc3/src/lib.rs#L51-L56) (link taken from commit). So either I'm wrong or glacier is wrong. Hmm.

pnkfelix:
Ah, Glacier invokes `rustc` directly. So maybe Glacier's logic is correct for direct `rustc` invocations, but not appropriate for `cargo` invocations?

Alexendoo:
Ah yeah sorry I didn't test cargo directly, I was using it with `--script rustc -- file` and assumed it would be the same

PR #79, opened 2020-05-03
Add ANSI color support and improved standard stream formatting

chrissimpkins:
The output from the tool is lengthy and it is difficult to view relevant information in the wall of text that streams by during execution. 

This PR adds ANSI escape code color support and a few std stream formatting improvements.

Nightly bisections:

Commit level bisections:

The final report and instructions for reporting:

Errors:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,6 +129,7 @@ name = "cargo-bisect-rustc"
 version = "0.4.1"
 dependencies = [
  "chrono 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "colored 1.9.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "dialoguer 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "dirs 2.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -205,6 +206,16 @@ version = "0.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bitflags 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "colored"
+version = "1.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "atty 0.2.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
Does this make copy/pasting the output into GitHub not work nicely? I seem to recall seeing boxes or whatever at some point appear when trying to do so.

chrissimpkins:
I didn't include ANSI escape codes in the report text intended for copy/paste onto the rust-lang/rust threads.  Here is the copy/paste of the text from the run in the screenshot above:

---

# Regression found in the compiler

searched nightlies: from nightly-2020-04-26 to nightly-2020-04-27
regressed nightly: nightly-2020-04-27
searched commits: from https://github.com/rust-lang/rust/commit/0862458dad90a0d80827e22e3f86e33add6d847c to https://github.com/rust-lang/rust/commit/e83f7563495dbe2629b0cbc738afb0808c4482e1
regressed commit: https://github.com/rust-lang/rust/commit/b592b37fa31c0d81ca23fea9fe7998c3f0f4bd0b
source code: URL OF A REPOSITORY THAT REPRODUCES THE ERROR

## Instructions

Please give the steps for how to build your repository (platform, system dependencies, etc.)
## Error

<details><summary>COLLAPSIBLE ERROR STACKTRACE</summary>
<p>

```bash
Paste the error the compiler is giving
```

</p></details>

---

Mark-Simulacrum:
Great, thanks! I'm on board with this though I don't have time to review the patch itself.

chrissimpkins:
As an aside, I don't ever see this full report pushed onto the repository.  It seems that most push this block of text:

```
searched nightlies: from nightly-2020-04-26 to nightly-2020-04-27
regressed nightly: nightly-2020-04-27
searched commits: from rust-lang/rust@0862458 to rust-lang/rust@e83f756
regressed commit: rust-lang/rust@b592b37
source code: URL OF A REPOSITORY THAT REPRODUCES THE ERROR
```

or just the regressed commit line and they manually add that this is coming from `bisect-rustc`.  Thoughts about changing the report to something like this:

---

searched nightlies: from nightly-2020-04-26 to nightly-2020-04-27
regressed nightly: nightly-2020-04-27
searched commits: from rust-lang/rust@0862458 to rust-lang/rust@e83f756
regressed commit: rust-lang/rust@b592b37
source code: URL OF A REPOSITORY THAT REPRODUCES THE ERROR

bisected with [cargo-bisect-rustc](https://github.com/rust-lang/cargo-bisect-rustc) v0.4.1

---

Mark-Simulacrum:
Seems reasonable. I think it's less well suited to Rust "users" (vs. the people doing triage), though -- I'm not sure whether it makes sense to prioritize that.

chrissimpkins:
It seems to me that the current workflow is generally:

- user reports the issue with a `rustc` stack trace +/- MCVE
- MCVE generated by reporting user or Cleanup Crew after triage requests it if it was not pushed in the initial report
- Someone (Core team, Cleanup Crew, the reporting user, someone who wanders by the thread and has an interest) runs `bisect-rustc` on the reduced or pre-reduced source and pastes the data into the thread

I haven't come across many cases when the stack trace, platform/env, and source used to reproduce are not available before the `bisect-rustc` report is pushed to a thread. This seems to be handled by those involved in diagnostics/triage rather than by users.

Mark-Simulacrum:
Yes, I realize that. Our original hope was that we could get the bisection working well enough that the original report would already contain the bisection (even if there's no MCVE prepared). But I think that hasn't actually materialized, so I wouldn't be opposed to optimizing for the more common case of cleanup crew doing bisection.

chrissimpkins:
Maybe something along the lines of the bot approach being discussed in https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/cargo-bisect.20as.20a.20bot would make it easier for users to run the tool as part of the error reporting process?

Mark-Simulacrum:
I think it's not going to pan out that users in practice will widely have the energy to learn our tooling, but the bot will make it much easier for people on the teams to run bisection, which is nice.

PR #88, opened 2020-05-14
Add cli test framework

pnkfelix:
Infrastructure for testing the command line tool.
    
Note: generates tests into fresh temp dir, which is deleted after testing is
 done (regardless of success or failure). You can change the
  `which_temp::WhichTempDir` type to revise this behavior.
    
This infrastructure includes two tests: `tests/cli.rs` and `tests/ice.rs`. Each
 uses very different strategies for testing cargo-bisect-rustc.
    
  1. `tests/cli.rs` uses a so-called meta-build strategy: the test inspects the
     `rustc` version, then generates build files that will inject (or remove,
     e.g. when testing `--regress=success`) `#[rustc_error]` from the source
     code based on the `rustc` version. This way, we get the effect of an error
     that will come or go based solely on the `rustc` version, without any
     dependence on the actual behavior of `rustc` itself (beyond its version
     string format remaining parsable).

     * This strategy should remain usable for the foreseeable future, without
       any need for intervention from `cargo-bisect-rustc` developers.
  
  2. `tests/ice.rs` uses a totally different strategy: It embeds an ICE that we
     know originated at a certain version of the compiler. The ICE is embedded
     in the file `src/ice/included_main.rs`. The injection point associated with
     the ICE is encoded in the constant `INJECTION_COMMIT`.

     * Over time, since we only keep a certain number of builds associated with
       PR merge commits available to download, the embedded ICE, the
       `INJECTION_COMMIT` definition, and the search bounds defined in
       `INJECTION_LOWER_BOUND` and `INJECTION_UPPER_BOUND` will all have to be
       updated as soon as the commit for `INJECTION_COMMIT` is no longer
       available for download.

     * Thus, this testing strategy requires regular maintenance from the
       `cargo-bisect-rustc` developers. (However, it is more flexible than the
       meta-build strategy, in that you can embed arbitrary failures from the
       recent past using this approach. The meta-build approach can only embed
       things that can be expressed via features like `#[rustc_error]`, which
       cannot currently express ICE's.

PR #94, opened 2020-05-24
Revert "Include compiler crashes in ICE defintion"

chrissimpkins:
This reverts the ICE definition change at commit a3891cdd26d1c5d35257c351c7c86fa7e72604bb.

Cherry picked https://github.com/rust-lang/cargo-bisect-rustc/pull/88/commits/7bd1ec404ff28377f2bfd2652fa86edea997f7e0 from #88 

r? @pnkfelix

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -240,33 +240,39 @@ impl Config {
             status, stdout_utf8, stderr_utf8
         );
 
-        // for commit message:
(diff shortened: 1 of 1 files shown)


pnkfelix:
Thanks for pulling this out!

I'll file an issue to track doing this the "right way." (According to the cargo devs, one can use verbose output from cargo which will print the underlying error codes from rustc, at least at *some* verbosity level...)

PR #96, opened 2020-05-26
Use multiple paths feature for cache step

imbsky:
Sorry, I missed this.


chrissimpkins:
Np! Thanks. 

And thanks for the updates to actions/cache!  I use it in a number of projects.  👍

imbsky:
I'm glad to hear that!

PR #98, opened 2020-06-13
Convert to default installation of cargo and replace `--with-cargo` option with `--without-cargo` option

### LINKED ISSUE #81: Mention `--with-cargo` in tutorial or make it a default
Without correct version of cargo any version of rustc that is older than few weeks will produce  random build errors which will be counted as "regression found" by default.

It is pretty confusing when `bisect-rustc` tells that revision `xyz` produces an error and silently stops bisecting correctly, while `cargo +xyz build` doesn't produce any error when run explicitly.

chrissimpkins:
Closes #81 

Also see the discussion in https://rust-lang.zulipchat.com/#narrow/stream/217417-t-compiler.2Fcargo-bisect-rustc/topic/unrecognized.20option.20json

This PR converts the `--with-cargo` optional behavior to default behavior to address  `Unrecognized option: 'json'` errors in tests as bisection proceeds back in time past a defined date (I believe that this is ~ July 2019'ish era).  The current default leads to incorrect automated detection of the regression commit due to this error.  These changes also remove the `--with-cargo` option.

Edit: @spastorino recommended the addition of the `--without-cargo` flag to disable the cargo install in https://github.com/rust-lang/cargo-bisect-rustc/pull/98#issuecomment-644284958.  This change was added in https://github.com/rust-lang/cargo-bisect-rustc/commit/2b7088c7f28fbf2dda7023a8fc30c092ab7764a1

### DIFF
--- a/src/toolchains.rs
+++ b/src/toolchains.rs
@@ -231,20 +231,21 @@ impl Toolchain {
             .map_err(InstallError::Download)?;
         }
 
-        // download nightly cargo by default
-        // this behavior was changed from an
-        // optional feature with the `--with-cargo`
-        // flag as of v0.6.0
+        // download cargo by default
+        // deactivate with the `--without-cargo` flag
+        // this default behavior was changed as of v0.6.0
         // see: https://github.com/rust-lang/cargo-bisect-rustc/issues/81
-        let filename = format!("cargo-nightly-{}", self.host);
-        download_tarball(
-            &client,
-            &format!("cargo for {}", self.host),
-            &format!("{}/{}/{}.tar", dl_params.url_prefix, location, filename,),
-            Some(&PathBuf::from(&filename).join("cargo")),
-            tmpdir.path(),
-        )
-        .map_err(InstallError::Download)?;
+        if !dl_params.without_cargo {
+            let filename = format!("cargo-nightly-{}", self.host);
+            download_tarball(
+                &client,
+                &format!("cargo for {}", self.host),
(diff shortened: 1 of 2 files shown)


spastorino:
Looks good but I wonder if we do want a `--without-cargo` kind of flag to turn this off.

chrissimpkins:
> Looks good but I wonder if we do want a `--without-cargo` kind of flag to turn this off.

Anyone opposed to this?  I'll add it this weekend if this is the only thing blocking the merge here.

chrissimpkins:
Rebased on master @ https://github.com/rust-lang/cargo-bisect-rustc/pull/98/commits/c5e86ee1c72f061bfbaaf80978737baa4bc1e19f

chrissimpkins:
`--without-cargo` flag added in 2b7088c7f28fbf2dda7023a8fc30c092ab7764a1

chrissimpkins:
@spastorino Please let me know if there is anything else that you need for this PR.  Thanks Santiago!

src/toolchains.rs
```diff
@@ -231,7 +231,11 @@ impl Toolchain {
             .map_err(InstallError::Download)?;
         }
 
-        if dl_params.install_cargo {
+        // download cargo by default
+        // deactivate with the `--without-cargo` flag
+        // this default behavior was changed as of v0.6.0
```
chrissimpkins, line 236: This line assumes a v0.6.0 release with the current changes if this PR is acceptable.

Issue #100, opened 2020-07-22
Adding the rustc-dev component


elichai:
Hi,
First, this is an awesome tool that I use frequently :) 
Second, I want to bisect something that uses `rustc_ast`, so for that I need also `rustc-dev` and currently it's not possible AFAIU.

Thanks!

PR #101, opened 2020-07-22
Add rustc-dev component option

elichai:
Fixes https://github.com/rust-lang/cargo-bisect-rustc/issues/100
I added a `--with-dev` option to download rustc-dev.
Because of https://github.com/rust-lang/rust/issues/72594 I also install llvm-tools together with rustc-dev

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -96,6 +96,9 @@ struct Opts {
     #[structopt(long = "with-src", help = "Download rust-src [default: no download]")]
     with_src: bool,
 
+    #[structopt(long = "with-dev", help = "Download rustc-dev [default: no download]")]
+    with_dev: bool,
+
     #[structopt(
         long = "test-dir",
         help = "Root directory for tests",
--- a/src/toolchains.rs
+++ b/src/toolchains.rs
@@ -259,6 +259,28 @@ impl Toolchain {
             .map_err(InstallError::Download)?;
         }
 
+        if dl_params.install_dev {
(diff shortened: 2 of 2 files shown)


ehuss:
Maybe it would be more flexible to add a `--component` flag similar to `rustup-toolchain-install-master` so that you can specify *any* component?

Mark-Simulacrum:
Yeah I think that would probably be the way to go, and we can still do some auto-detection for installing llvm-tools-preview.

It might also be good to consider dropping our custom installation code in favor of forking out to rustup-toolchain-install-master, and documenting that as a pre-req for bisection. (Maybe we can get a library version of it published, too).

spastorino:
What is the status of this? cc @elichai

elichai:
@spastorino not sure what's the next step, can it be merged as is? or do we want the `--component` flag? or maybe depend on `rustup-toolchain-install-master` somehow?

spastorino:
Maybe @ehuss or @Mark-Simulacrum can clarify a bit and guide you better than I

Mark-Simulacrum:
I think we should probably just merge. While I continue to think that a general component flag is the way to go, this is still likely a useful step. Ultimately I do think it makes sense for us to find a way to avoid a bunch of tools repeating the rustup-toolchain-install-master-like logic to install a set of rustup components for a particular commit; perhaps that means rustup-toolchain-install-master should become a library instead of just a cargo-install command, not sure.

PR #106, opened 2020-09-09
Document how to build for musl host

lzutao:
to help statically builds with musl toolchain.


bjorn3:
`cargo build --features git2/vendored-openssl` should already work before this PR.

lzutao:
TIL, thanks. But I think the discovery of that is lower than direct feature flags.

spastorino:
I think I agree with @bjorn3 here and would close this PR but maybe we can document somewhere (on the README?) @bjorn3 solution?.

PR #107, opened 2020-09-09
Add support for RUST_SRC_URL var at runtime

### LINKED ISSUE #104: why RUST_SRC_REPO needs to know at compile time ?
https://github.com/rust-lang/cargo-bisect-rustc/blob/843b96ec3dfa26e7c7d51d3dd589c2f8f0705cae/src/git.rs#L11

If I build cargo-bisect-rustc and copy it to another computer with different directory to rustc,
does it make me slowly re-clone rustc ?

lzutao:
With this PR, now the ordering to find rustc repo path is:
* RUST_SRC_URL at runtime
* rustc.git in current dir at runtime
* RUST_SRC_URL set at compilation time

Closes #104.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -3,6 +3,7 @@
 const RUST_SRC_URL: &str = "https://github.com/rust-lang/rust";
 const RUST_SRC_REPO: Option<&str> = option_env!("RUST_SRC_REPO");
 
+use std::env;
 use std::path::Path;
 
 use chrono::{TimeZone, Utc};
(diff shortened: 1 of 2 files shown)


src/git.rs
```diff
@@ -39,26 +40,33 @@ fn lookup_rev<'rev>(repo: &'rev Repository, rev: &str) -> Result<Git2Commit<'rev
...
-            Ok(repo)
+    fn open(repo: &Path) -> Result<Repository, Error> {
+        eprintln!("opening existing repository at {:?}", repo);
+        let repo = Repository::open(repo)?;
+        {
+            eprintln!("refreshing repository");
+            // Sometimes origin point to github fork of contributor,
+            // let's inspect upstream remote first.
+            let mut remote = repo
+                .find_remote("upstream")
```
spastorino, line 51: I understand what you're trying to do here but wouldn't it be better to pass this also as an env var?.

spastorino (commented):
Seems good to me, the only thing I'm not sure about is the `upstream` idea because unsure if everyone uses the same criteria here. Maybe it's better to pass the branch as an env var as people may also want to point to their rust checkout under their feature branch.

src/git.rs
lzutao: I am think about using `remote_anonymous` with rustc URL instead.
This has the advantages that we don't have to guess the remote names the and their URLs.
Another problem is sometime the remote URL is alias for ssh host, does git2/we handle it.

lzutao:
I moved the upstream related changes to other PRs.

PR #109, opened 2020-09-11
Anonymously fetch from upstream rust repo

lzutao:
`origin` or `upstream` are commonly used remote names. 
However they aren't always the right remote with URL points to rustc URL.
Their URLs could contain SSH aliases that we don't resolve.

Let's fetch anonymously from https://github.com/rust-lang/rust and use FETCH_HEAD
to refer to it.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -38,9 +38,7 @@ fn get_repo() -> Result<Repository, Error> {
         let repo = Repository::open(repo)?;
         {
             eprintln!("refreshing repository");
-            let mut remote = repo
-                .find_remote("origin")
-                .or_else(|_| repo.remote_anonymous("origin"))?;
+            let mut remote = repo.remote_anonymous(RUST_SRC_URL)?;
             remote.fetch(&["master"], None, None)?;
         }
         Ok(repo)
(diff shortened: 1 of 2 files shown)


lzutao:
Related small discussion: https://github.com/rust-lang/cargo-bisect-rustc/pull/107#discussion_r486762009

Mark-Simulacrum:
Hm, I think historically I've found that we shouldn't need an explicit remote name and can instead directly use the https:// URL instead of the "name" of a remote. Is that not the case here?

lzutao:
I thought about that. Fetching directly from https:// save states in local repo 
as FETCH_HEAD, maybe we could use that instead of origin/master ?

Mark-Simulacrum:
If that works, then I think that would be better -- avoiding dropping our state into people's repositories is good.

lzutao:
Done. But I haven't had a change to run to FETCH_HEAD logic in `AccessViaLocalGit::commits`.

spastorino:
@lzutao after merging #108 this one has conflicts now.

lzutao:
Rebased.

spastorino:
@Mark-Simulacrum any other thought before merging?

Mark-Simulacrum:
Seems fine to me.

PR #116, opened 2020-10-08
Calculate current nightly date only once

lzutao:
This function (`default_nightly`) is hot and called repeatedly.
It also calls external rustc and compiles a regex.


lzutao:
It's been two weeks without reviews :| .

apiraino:
@lzutao apologies, your 4 PRs perhaps passed under the radar. Try pinging `@spastorino` on all of them

r? @spastorino

PR #118, opened 2020-10-21
Release v0.5.3

### LINKED ISSUE #117: Release new version
**TL;DR** Latest official version cannot bisect past circa May – I would appreciate a new release

In May, **`@alexcrichton`** introduced the embed-bitcode option to Rust. Unfortunately, the latest official version of cargo-bisect-rustc uses the **system cargo** (for me, the latest version of cargo) to invoke rustc, which means that anytime I bisect longer ago than May, I get an error from rustc saying it doesn't recognize the option. @ehuss said in Zulip that PR #98 fixes this, but it would be nice to be able to download an official released version with `cargo install`.

**EDIT:** This should have a major version bump, so it will be 0.6.0, not 0.5.3.

camelid:
Fixes #117.

r? @spastorino

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.5.3
+
+### Added
+
+- Support specifying the path to a rust-lang/rust clone at runtime with `RUST_SRC_REPO`
+
+### Changed
+
+- Make `--with-cargo` the default to allow bisecting past changes in rustc options. Add `--without-cargo` flag to use the old behavior.
+- Use an anonymous remote that always points to rust-lang/rust when refreshing repository
+
+### Fixed
+
+- Add nightly start and end date validations against the current date – previously would attempt to install nightly even if date was in the future
+- Verify that `--test-dir` is a directory instead of assuming it is and then panicking
+
 ## v0.5.2
 
 - Fix: revert the revised internal compiler error definition in commit a3891cdd26d1c5d35257c351c7c86fa7e72604bb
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
 
+## v0.5.3
+
+### Added
+
+- Support specifying the path to a rust-lang/rust clone at runtime with `RUST_SRC_REPO`
+
+### Changed
+
+- Make `--with-cargo` the default to allow bisecting past changes in rustc options. Add `--without-cargo` flag to use the old behavior.
```
camelid, line 11: I put these on a single line even though I think that might be a "no-no" because they were part of the same change.

camelid:
Actually, do we consider #98 to be a breaking change?

Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 name = "cargo-bisect-rustc"
 readme = "README.md"
 repository = "https://github.com/rust-lang/cargo-bisect-rustc"
-version = "0.5.2"
```
spastorino, line 11: Let's make it `0.6.0` given that more or less it changes some behavior.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.5.3
+
+### Added
```
spastorino, line 5: I'd keep the same style we were using so I'd avoid sub-titles and just go with a plan list here.

camelid: I was trying to follow Keep A Changelog since that's what we use for other projects. I also find the subheads more readable than having them inline. But I won't block on that, so let me know if I should still change it :)

camelid:
Closed in favor of #119.

PR #119, opened 2020-10-22
Release v0.6.0

### LINKED ISSUE #117: Release new version

camelid:
Fixes #117. Supersedes #118.

r? @spastorino

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.6.0
+
+### Added
+
+- Support specifying the path to a rust-lang/rust clone at runtime with `RUST_SRC_REPO`
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


spastorino:
I've just released a new version!

PR #121, opened 2020-11-01
Improve message for no CI builds found and various cleanup

camelid:
- Improve message for no CI builds found
- Various cleanup


src/main.rs
```diff
@@ -1146,7 +1142,7 @@ fn bisect_ci_in_commits(
 
     if commits.is_empty() {
         bail!(
-            "no commits between {} and {} within last 167 days",
+            "no CI builds available between {} and {} within last 167 days",
```
camelid, line 1145: This error message confuses me: What does it mean by "within last 167 days"?

spastorino: after 167 days artifacts are removed from the CI, so you can't install older nightlies.

PR #130, opened 2021-05-19
Use git CLI to fetch a repo.

ehuss:
This changes the git fetch of the rust-lang/rust repo to use the `git` CLI command instead of libgit2. This is because libgit2 has some performance issues with large repos. On my system, a fetch that is not particularly old takes about 4 minutes with libgit2, whereas with the git CLI it takes about 2 seconds.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -7,7 +7,7 @@ use std::env;
 use std::path::Path;
 
 use chrono::{TimeZone, Utc};
-use failure::{bail, Error};
+use failure::{bail, Error, ResultExt};
 use git2::build::RepoBuilder;
 use git2::{Commit as Git2Commit, Repository};
 use log::debug;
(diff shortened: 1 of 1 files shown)


src/git.rs
```diff
@@ -34,13 +34,20 @@ fn lookup_rev<'rev>(repo: &'rev Repository, rev: &str) -> Result<Git2Commit<'rev
 
 fn get_repo() -> Result<Repository, Error> {
     fn open(repo: &Path) -> Result<Repository, Error> {
+        eprintln!("refreshing repository at {:?}", repo);
+        // This uses the CLI because libgit2 is quite slow to fetch a large repository.
+        let status = std::process::Command::new("git")
+            .arg("fetch")
```
camelid, line 40: I think this needs to use something like the old `repo.remote_anonymous(RUST_SRC_URL)`. E.g., if I run `git fetch` in my rust-lang/rust clone, it only fetches my fork, which doesn't have the latest changes.

ehuss: Good point!  I normally point it at a repo where origin is rust-lang/rust, but it makes sense to use the correct remote.  I don't think it can use an anonymous remote (since that is in memory only), but I added the URL to the fetch line which should do the right thing.

PR #132, opened 2021-05-29
Fix typos in the tutorial

LingMan:


hellow554:
Those are not really typos. You can call both. I for example do `cargo-bisect-rustc` so I can find it quicker in my fish history ;)

LingMan:
Doesn't work for me:
```
$ cargo-bisect-rustc
bash: cargo-bisect-rustc: command not found
```
```
$ cargo bisect-rustc
installing nightly-2021-07-12
[...]
```

Also this file currently has three instances of calling  `cargo bisect-rustc` and three instances of calling `cargo-bisect-rustc`, so there's at least a consistency problem.

hellow554:
> Edit: This doesn't work for me because the cargo-bisect-rustc binary gets installed into ~/.cargo/bin, which is not in my PATH.

`~/.cargo/bin` should be in your path when you want to use executables installed by `cargo install`.

It's interesting to see, that `cargo bisect-rustc` does work for you. Not sure if cargo directly looks into the bin directory without regarding the `$PATH` (which at least seems so)

LingMan:
> `~/.cargo/bin` should be in your path when you want to use executables installed by `cargo install`.

That may make semantic sense for arbitrary executables, but cargo subcommands should and do work without `PATH`. That view is supported by the fact that cargo [indeed explicitly searches](https://github.com/rust-lang/cargo/blob/38c8ce6cabb29a4261a021896ce34f09d1c16189/src/bin/cargo/main.rs#L192) the configured bin directory (in addition to `PATH`).

Now, is `cargo-bisect-rustc` a cargo subcommand? I'd argue that, yes, it is, because it follows the established naming scheme of `cargo-%command_name%` and operates on cargo projects.

Additionally, as I already pointed out, TUTORIAL.md is currently inconsistent between calling `cargo bisect-rustc` and `cargo-bisect-rustc`. Resolving that inconsistency alone should be grounds for merging this PR (and is why I titled it as a typo fix).

oli-obk:
Yea, I prefer the subcommand, too

Issue #84, opened 2020-05-09
Add timeout


bjorn3:
To bisect hangs.

Mark-Simulacrum:
Timeout to what? What hangs? Can you provide more clarification?

bjorn3:
If the rustc regression is a hang, it would be useful for cargo-bisect-rustc to timeout after testing for a certain amount of time and then considering that timeout as if the compiler crashed.

bjorn3:
For context: I was trying to bisect https://github.com/rust-lang/rust/issues/72051.

hkratz:
Quite a few "extreme" slowness / hang issues lately where that would help do an unattended bisect instead of using `--prompt`.

ehuss:
This is now implemented via #135 and should be in the recently published 0.6.1.  Feel free to open new issues if there are any problems with the new `--timeout` option.

PR #135, opened 2021-10-04
add timeout functionality for bisecting hangs 

dvente:
This is in response to #84. In addition to all the tests still running, I've manually tested it by using it trying to bisect rust-lang/rust#75860 and it seemed to work fine there. Seeing as it's a very small change I'm assuming that that is sufficient, but if more is needed, please let me know. 

Couple of things to mention: 
- I have not tested the code on windows as I don't have an env for that, but given that the code uses the `std::process::Command` and Windows also has a similar functinality I'm assuming that that will be sorted out automatically
- I'm not a very experienced rust dev so the clones etc. in line 340-342 are probably not necessary but I couldn't figure out how to do it without. If someone knows a better way, feedback is welcome. 
Hope this helps, If there is anything that's missing, please let me know!


src/toolchains.rs
```diff
@@ -334,6 +335,31 @@ impl Toolchain {
                 }
                 cmd
             }
+            (Some(script), Some(timeout)) => {
+                let mut cmd = Command::new("timeout");
+                let mut args: Vec<std::ffi::OsString> =
+                    vec![script.into(), format!("{}", timeout).into()];
```
12101111, line 341: The order is reversed.

```suggestion
                    vec![format!("{}", timeout).into(), script.into()];
```

12101111 (commented):
Thanks for this PR. It works fine for me, expect the typo here.

src/toolchains.rs
```diff
@@ -334,6 +335,31 @@ impl Toolchain {
...
+                let mut args: Vec<std::ffi::OsString> =
+                    vec![format!("{}", timeout).into(), script.into()];
+                //FIXME: be smarter so we don't have to clone
+                args.extend(cfg.args.command_args.clone());
+                cmd.env("RUSTUP_TOOLCHAIN", self.rustup_name());
+                cmd.args(args);
+                cmd
+            }
+            (None, Some(timeout)) => {
+                // let mut cmd = Command::new("cargo");
```
spastorino, line 349: Remove this?

spastorino (approved):

spastorino:
Leaving this up to @ehuss

PR #140, opened 2021-12-07
Add `--regress` example to tutorial.

orium:
~The script must exit with 0 if there is not problem and non-zero otherwise.  Since `grep` exits with 0 if a line matches the pattern we need to negate its exit code.~

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -89,7 +89,10 @@ cargo bisect-rustc --test-dir=foo --start=2018-05-07 --end=2018-08-04
 ```
 
 By default it will run `cargo build` in the project and check whether or not
-it fails. In just a few steps, we find that it stopped working on
+it fails.  You can also use the flag `--regress` to specify other common
+regression criteria, e.g. `--regress=ice` for internal compiler errors.
+
+In out example, in just a few steps, we can we find that it stopped working on
 `nightly-2018-07-30`.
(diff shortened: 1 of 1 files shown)


ehuss:
I believe the original is correct.  It is showing how to automate the previous example, which was trying to detect when `E0642` *stopped* being produced.  That is, emitting E0642 is "good" and not emitting it is "bad".  

Perhaps you can add a short comment to the script explaining what it is doing?

orium:
Maybe I can give a more clear example that doesn't require explanation (and also reflect a common case):

```bash
! cargo check 2>&1 | grep 'internal compiler error'
```

What do you think?

ehuss:
I probably wouldn't include that as a specific example as it wouldn't be something we'd expect someone to ever write. Detecting ice's is already built in (via `--regress=ice`).  Scripts are best suited for unusual situations, and I think illustrating a real-world example (examining issue https://github.com/rust-lang/rust/issues/55036) is a good way of demonstrating what a script would actually be used for.

orium:
> I probably wouldn't include that as a specific example as it wouldn't be something we'd expect someone to ever write.

I did :slightly_smiling_face:.  Didn't know about `--regress`, so I've added an example to the tutorial.

ehuss (approved):
Thanks!

PR #142, opened 2022-02-09
Include rustc overflowing its stack as an ICE

Alexendoo:
Helpful for when you're bisecting something that overflows before emitting an ICE diagnostic


ehuss (approved):
I'm a bit on the fence, as I don't usually equate an overflow with ICE. But since they are pretty similar, and I doubt anyone needs to check them independently, I'm going to go ahead and merge.  Thanks!

PR #147, opened 2022-03-30
Add support for git tags in `--start` and `--end`

Kobzol:
First I tried to add new `Bound` (`Version`), but that required a lot of changes in the code.

I think that just adding support for arbitrary tags is more general and should hopefully work out of the box. I changed the GH resolver to use the `compare` API, which allows us to get commit information together with `merge-base` ancestor.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -26,7 +26,21 @@ impl Commit {
 }
 
 fn lookup_rev<'rev>(repo: &'rev Repository, rev: &str) -> Result<Git2Commit<'rev>, Error> {
-    if let Ok(c) = repo.revparse_single(rev)?.into_commit() {
+    let revision = repo.revparse_single(rev)?;
+
+    // Find the merge-base between the revision and master.
+    // If revision is a normal commit contained in master, the merge-base will be the commit itself.
+    // If revision is a tag (e.g. a release version), the merge-base will contain the latest master
+    // commit contained in that tag.
+    let master_id = repo.revparse_single("origin/master")?.id();
+    let revision_id = revision
+        .as_tag()
+        .map(|tag| tag.target_id())
+        .unwrap_or_else(|| revision.id());
+
+    let common_base = repo.merge_base(master_id, revision_id)?;
+
+    if let Ok(c) = repo.find_commit(common_base) {
         return Ok(c);
     }
     bail!("Could not find a commit for revision specifier '{}'", rev)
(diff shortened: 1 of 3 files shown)


ehuss:
Nice!

Can you add some text to the CLI help for the `--start` and `--end` flags that documents the kinds of values they support?  Maybe something like `[a YYYY-MM-DD date or git tag or nightly git SHA]` at the end?

Kobzol:
Added CLI help. Maybe we should also rename `Bound::Commit` to something like `Bound::GitReference`?

Kobzol:
Done.

ehuss (approved):
Thanks!

ehuss:
@Kobzol This seems to have broken bisecting by hash.  Can you take a look at fixing it?  For example:

`cargo-bisect-rustc --start e08d5693609a659e45025b8ea4dbd9efa342fa68 --end ec667fbcfcd29e2b5080f6be04e229356b27e600`

will error with:

```
bisecting ci builds
starting at e08d5693609a659e45025b8ea4dbd9efa342fa68, ending at ec667fbcfcd29e2b5080f6be04e229356b27e600
refreshing repository at "/Users/eric/Proj/rust/rust-upstream"
From https://github.com/rust-lang/rust
 * branch                    HEAD       -> FETCH_HEAD
opening existing repository at "/Users/eric/Proj/rust/rust-upstream"
fetching (via local git) commits from e08d5693609a659e45025b8ea4dbd9efa342fa68 to 175b8db73bfd078b4bcd3c28c8d6f51d5895ebf3
refreshing repository at "/Users/eric/Proj/rust/rust-upstream"
From https://github.com/rust-lang/rust
 * branch                    HEAD       -> FETCH_HEAD
opening existing repository at "/Users/eric/Proj/rust/rust-upstream"
looking up first commit
looking up second commit
checking that commits are by bors and thus have ci artifacts...
finding bors merge commits
ERROR: failed during attempt to create/access local git repository: Expected author Brian Anderson to be bors for beb9a0dfc52ebda4f8db4e5d439e08e4f3a43a39.
 Make sure specified commits are on the master branch!
```

I notice that the end commit has shifted from ec667fbcfcd29e2b5080f6be04e229356b27e600 to
175b8db73bfd078b4bcd3c28c8d6f51d5895ebf3.

Also, if cargo-bisect-rustc is compiled to clone a bare `rust.git` repo (the default), it simply fails to start with:

Kobzol:
It seems that the problem is that `origin/master` is not updated when performing `merge-base`. A short workaround is to delete the `rust.git` directory (or `rust-upstream` in your case). I will send a PR that will update `origin/master` before calculating `merge-base` ancestor.

PR #149, opened 2022-04-16
Fetch newest origin/master before calculating revision base

Kobzol:
Fixes https://github.com/rust-lang/cargo-bisect-rustc/pull/147#issuecomment-1100273375

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -28,6 +28,9 @@ impl Commit {
 fn lookup_rev<'rev>(repo: &'rev Repository, rev: &str) -> Result<Git2Commit<'rev>, Error> {
     let revision = repo.revparse_single(rev)?;
 
+    // Update origin/master, since it might be stale
+    repo.find_remote("origin")?.fetch(&["master"], None, None)?;
+
     // Find the merge-base between the revision and master.
     // If revision is a normal commit contained in master, the merge-base will be the commit itself.
     // If revision is a tag (e.g. a release version), the merge-base will contain the latest master
--- a/src/git.rs
+++ b/src/git.rs
@@ -4,6 +4,7 @@ const RUST_SRC_URL: &str = "https://github.com/rust-lang/rust";
 const RUST_SRC_REPO: Option<&str> = option_env!("RUST_SRC_REPO");
 
 use std::env;
+use std::ops::Deref;
 use std::path::Path;
 
 use chrono::{TimeZone, Utc};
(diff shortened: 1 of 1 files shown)


Kobzol:
CC @ehuss

ehuss:
Thanks for taking a look so quickly!

Unfortunately this doesn't seem to resolve the issue. There are a few problems with this:

* `origin` is my fork of rust-lang/rust on GitHub, which may be out of date
* libgit2 can't fetch from `origin` because it is an SSH URL, so it can't authenticate
* I tried fetching using the CLI, and it didn't seem to help
* Fetching with libgit2 can be extremely slow.  cargo-bisect-rustc uses the git CLI to fetch [here](https://github.com/rust-lang/cargo-bisect-rustc/blob/47d736ad955153f63ad513f1f2aaed595eaa213a/src/git.rs#L52-L60) for performance reasons.

I realize looking at the original PR, it seems to hard-code `origin/master`.  `origin` may not be `RUST_SRC_URL`. The correct remote name for my setup is `upstream`.

I'm wondering if it would be possible to determine the correct remote name by iterating over all the remotes and searching for one that has RUST_SRC_URL as the url?

Kobzol:
I wonder if it would be easier to just add a new remote automatically (with some uncommon name, like `__bisect_origin_fetch`) and use that one?

Kobzol:
I changed the logic to find the origin branch and fetch it when opening the repository.

src/git.rs
```diff
@@ -46,36 +62,55 @@ fn lookup_rev<'rev>(repo: &'rev Repository, rev: &str) -> Result<Git2Commit<'rev
...
+    })
+}
+
+fn find_origin_remote(repo: &Repository) -> Result<String, Error> {
+    repo.remotes()?
+        .iter()
+        .filter_map(|name| name.and_then(|name| repo.find_remote(name).ok()))
+        .find(|remote| remote.url().map(|url| url.contains(RUST_SRC_URL)).unwrap_or(false))
+        .and_then(|remote| remote.name().map(|name| name.to_string()))
+        .ok_or_else(|| failure::err_msg("Origin remote branch not found. Try to add a HTTPS remote pointing to the main rustc repository."))
```
ehuss, line 113: I think it might be good to include the actual URL the user should add, and where they should add it.  Also, I think the use of the term "origin" here could be confusing, since "origin" could mean different things. Perhaps something like this?

```suggestion
        .ok_or_else(|| {
            failure::format_err!(
                "rust-lang/rust remote not found. \
                 Try adding a remote pointing to {} in the rust repository at {}.",
                RUST_SRC_URL,
                repo.path().display()
            )
        })
```

ehuss (commented):
Thanks! It looks like this should probably work.

It looks like the code needs to be formatted.

Kobzol:
I have adapted the code based on your suggestion :)

Badel2:
Thank you, I was about to open an issue because I also got the error "Expected author Brian Anderson to be bors", which is not an acceptable error message.

Automatic fetching is a nice solution, can't wait for the next release!

ehuss (approved):
Thanks!

PR #152, opened 2022-05-04
Code refactor

kraktus:
No functional changes

As [posted on zulip](https://rust-lang.zulipchat.com/#narrow/stream/217417-t-compiler.2Fcargo-bisect-rustc) this part of a larger effort to modernise the code of the project, but for the code review to be easier thought it would be better to split it in parts. The PR contains quite a few commits compared to the changes but each of those is a logic unit that pass the tests and with formatted code.


src/main.rs
```diff
@@ -94,10 +86,7 @@ struct Opts {
     #[structopt(long = "preserve", help = "Preserve the downloaded artifacts")]
     preserve: bool,
 
-    #[structopt(
-        long = "preserve-target",
-        help = "Preserve the target directory used for builds"
-    )]
+    #[structopt(long, help = "Preserve the target directory used for builds")]
     preserve_target: bool,
```
oli-obk, line 95: are underscores converted to dashes?

oli-obk (approved):
just one question that wasn't clear to me from the documentation of clap/structop

src/main.rs
kraktus: Yes it does, `-h` after the changes


````
cargo-bisect-rustc 0.6.3

USAGE:
    cargo-bisect-rustc [FLAGS] [OPTIONS] [-- <command-args>...]

FLAGS:
    -a, --alt                Download the alt build instead of normal build
        --by-commit          Bisect via commit artifacts
        --force-install      Force installation over existing artifacts
    -h, --help               Prints help information
        --preserve           Preserve the downloaded artifacts
        --preserve-target    Preserve the target directory used for builds
        --prompt             Manually evaluate for regression with prompts
    -V, --version            Prints version information
        --verbose            
        --with-dev           Download rustc-dev [default: no download]
        --with-src           Download rust-src [default: no download]
        --without-cargo      Do not install cargo [default: install cargo]

OPTIONS:
        --access <access>              How to access Rust git repository [github|checkout]
    -c, --component <components>...    additional components to install
        --end <end>                    Right bound for search (*with* regression). You can use a date (YYYY-MM-DD), git
                                       tag name (e.g. 1.58.0) or git commit SHA.
        --host <host>                  Host triple for the compiler [default: unknown]
        --install <install>            Install the given artifact
        --regress <regress>            Custom regression definition [default: error]
        --script <script>              Script replacement for `cargo build` command
        --start <start>                Left bound for search (*without* regression). You can use a date (YYYY-MM-DD),
                                       git tag name (e.g. 1.58.0) or git commit SHA.
        --target <target>              Cross-compilation target platform
        --test-dir <test-dir>          Root directory for tests [default: .]
    -t, --timeout <timeout>            Assume failure after specified number of seconds (for bisecting hangs)

ARGS:
    <command-args>...    Arguments to pass to cargo or the file specified by --script during tests

EXAMPLES:
    Run a fully automatic nightly bisect doing `cargo check`:
    ```
    cargo bisect-rustc --start 2018-07-07 --end 2018-07-30 --test-dir ../my_project/ -- check
    ```

    Run a PR-based bisect with manual prompts after each run doing `cargo build`:
    ```
    cargo bisect-rustc --start 6a1c0637ce44aeea6c60527f4c0e7fb33f2bcd0d \
      --end 866a713258915e6cbb212d135f751a6a8c9e1c0a --test-dir ../my_project/ --prompt -- build
    ```
````

PR #153, opened 2022-05-13
Reduce technical debt

kraktus:
No functional change

Replace deprecated/no-longer-maintained crates with a up-to-date replacement:

- `structopt` -> `clap` allowing for more validation to be done during arguments passing
- `failure` -> `thiserror` + `anyhow`

along with more tests and some refactoring.

this PR has maybe grown too big, and I can split it if needed, although efforts were made for it to be reviewed on a per-commit basis.


oli-obk:
General note: I think I would prefer `eyre` (with `color-eyre`) over `anyhow`, but that's not a strong requirement

Cargo.toml
```diff
@@ -34,3 +34,4 @@ colored = "2"
 
 [dev-dependencies]
 quickcheck = "0.9.2"
+trycmd = "0.13.4"
```
oli-obk, line 38: seems odd that it creates empty files for empty stderr, maybe patch that upstream?

src/main.rs
```diff
@@ -904,10 +904,8 @@ fn bisect_nightlies(cfg: &Config, client: &Client) -> anyhow::Result<BisectionRe
                         "the start of the range ({}) must not reproduce the regression",
                         t
                     );
-                } else {
-                    last_failure = nightly_date;
                 }
-
+                last_failure = nightly_date;
```
oli-obk: this changes behaviour

src/main.rs
```diff
@@ -478,6 +479,403 @@ impl Config {
             target,
             toolchains_path,
             rustup_tmp_path,
+            client: Client::new(),
+        })
+    }
+
+    fn install(&self, bound: &Bound) -> anyhow::Result<()> {
```
oli-obk, line 486: this commit is very noisy, maybe create impl blocks wherever there were functions before? It's ok to have multiple impl blocks

src/main.rs
kraktus: I don't think it does,  if any conditional branch is activated, this line is not reached. Was caught by clippy.

Cargo.toml
kraktus: The documentation states that if the files are not present the output is not checked. tested and it behaves like that. I don't think those empty files are an issue

oli-obk: Ah... unfortunate difference to rustc's ui tests. 🤷

src/main.rs
kraktus: thanks modified the commit to make it less noisy

kraktus:
CI fixed, and previous comments should have been addressed.

> General note: I think I would prefer `eyre` (with `color-eyre`) over `anyhow`, but that's not a strong requirement

I am not familiar with the library, from what I see the difference is that it includes `spantrace`. It will be a bit more noisy but considering our target are developers it should not be an issue. If we're good with leaving it by default, can make the switch.

oli-obk:
Thanks!

I went ahead and merged it, since it can stand on its own.

Yea, using spantrace and being more developer focussed is good here, as failures in the tool are likely going to be debugged by users of the tool as you noted. Let's do that in a separate PR tho

Issue #154, opened 2022-05-24
Segmentation fault when cloning rust repository on systems with libgit2 >= 1.4.0


LingMan:
TLDR: git2 dependency update needed. Will post a PR shortly, this issue is just for the record.

Reproduction:
- Ensure you have libgit2 version 1.4.0 or greater installed on your system (as is currently the case with Arch and Manjaro)
- Run `cargo install cargo-bisect-rustc`. Do NOT set the `RUST_SRC_REPO` environment variable.
- Prepare a crate to run a successful bisection with. For example, put this into `main.rs` (regression in nightly-2022-04-17):
```rust
fn f() {
    g((), ());
}

pub fn g(
    a1: (),
    a2: bool,
    a3: bool,
    a4: bool,
    a5: bool,
    a6: ()
) -> () {
}

fn main() {
}
```
- Run a successful bisection. Do NOT set the `RUST_SRC_REPO` environment variable here either. If you use the example from above run
`cargo bisect-rustc --regress=ice --start=2022-04-16 --end=2022-04-17`
The last output before the segfault should be `cloning rust repository`.
The segfault itself happens at this line:
https://github.com/rust-lang/cargo-bisect-rustc/blob/51b048455074007af6082918d065701e34e83935/src/git.rs#L95

The stacktrace for me was:
```
Stack trace of thread 170034:
                #0  0x00007f5edef15260 n/a (libgit2.so.1.4 + 0xa4260)
                #1  0x00007f5edef16d6e git_remote_fetch (libgit2.so.1.4 + 0xa5d6e)
                #2  0x00007f5edeea20ac n/a (libgit2.so.1.4 + 0x310ac)
                #3  0x000055ef51df2d15 n/a (/home/username/.cargo/bin/cargo-bisect-rustc + 0x18bd15)
```

Thankfully the problem was already reported to git2 in https://github.com/rust-lang/git2-rs/issues/813 and fixed. We just need to update the dependency. As mentioned at the beginning, I'll submit a PR shortly.

PR #155, opened 2022-05-24
Update dependencies

LingMan:
This updates all dependencies to current versions.

The first commit fixes https://github.com/rust-lang/cargo-bisect-rustc/issues/154.

Note that `tempdir` is unmaintained and should be replaced with `tempfile`, but that's a task for another day and PR.


oli-obk:
Nice! Maybe we should have a cronjob to check for relevant updates

PR #159, opened 2022-06-16
set `CARGO_BUILD_TARGET` in bisect run

Emilgardis:


Emilgardis:
related to cross-rs/cross#805 and cross-rs/cross#669, this should help with some pitfalls when using `cargo-bisect-rustc` and a script while setting a target

PR #177, opened 2022-07-28
Allow tag-like bounds that are older than 167 days.

### LINKED ISSUE #167: Use dates for git tags if possible
Support for using git tags in `--start` and `--end` was added in #147. cargo-bisect-rustc translates those to hashes. However, CI commits or only kept for a few months (167 days?). That means you can't go back very far using git tags. I think it would be better if it translated those tags into dates, which go back forever, allowing bisecting of older releases.  

If the range is within the last 167 days, cargo-bisect-rustc will automatically switch to CI commits once it has found the specific nightly.

Translating a commit to a nightly date isn't straightforward. However, I think it can use a heuristic of assuming that the commit exists in the nightly of the next day after the commit.  That's not always true, but I think should be close enough.

ehuss:
When using a tag bound like `--start 1.53.0`, it would try to find the SHA commit that is associated with that tag and then bisect between commits. However, since commits don't go back very far, this means that tagged bounds only support roughly the last 5 releases. I often need to bisect older versions. This changes it so that when given a tag, it will translate that to a nightly date instead.

Fixes #167

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -484,6 +484,42 @@ impl Config {
     }
 }
 
+/// Translates a tag-like bound (such as `1.62.0`) to a `Bound::Date` so that
+/// bisecting works for versions older than 167 days.
+fn fixup_bounds(
+    access: &Access,
+    start: &mut Option<Bound>,
+    end: &mut Option<Bound>,
+) -> anyhow::Result<()> {
+    let is_tag = |bound: &Option<Bound>| -> bool {
+        match bound {
+            Some(Bound::Commit(commit)) => commit.contains('.'),
+            None | Some(Bound::Date(_)) => false,
+        }
+    };
+    let is_datelike = |bound: &Option<Bound>| -> bool {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -484,6 +484,42 @@ impl Config {
...
+    if !(is_datelike(start) && is_datelike(end)) {
+        // If the user specified an actual commit for one bound, then don't
+        // even try to convert the other bound to a date.
+        return Ok(());
+    }
+    let fixup = |bound: &mut Option<Bound>| -> anyhow::Result<()> {
+        if is_tag(bound) {
+            if let Some(b) = bound.clone() {
+                if is_tag(bound) {
+                    *bound = Some(Bound::Date(access.repo().bound_to_date(b)?));
```
oli-obk, line 512: Maybe print a message here saying that this happened. That seems like a good thing for the user to know about.

ehuss: Sounds good!

Issue #85, opened 2020-05-12
Nightly can be checked thrice in certain cases


hellow554:
```
cargo-bisect-rustc --end=2020-05-09 --script ./test.sh --access github
installing nightly-2020-05-09
RESULT: nightly-2020-05-09, ===> Yes
uninstalling nightly-2020-05-09

installing nightly-2020-05-07
RESULT: nightly-2020-05-07, ===> Yes
uninstalling nightly-2020-05-07

installing nightly-2020-05-05
RESULT: nightly-2020-05-05, ===> Yes
uninstalling nightly-2020-05-05

installing nightly-2020-05-03
RESULT: nightly-2020-05-03, ===> No
uninstalling nightly-2020-05-03

installing nightly-2020-05-05
RESULT: nightly-2020-05-05, ===> Yes
uninstalling nightly-2020-05-05

installing nightly-2020-05-04
RESULT: nightly-2020-05-04, ===> No
uninstalling nightly-2020-05-04

searched toolchains nightly-2020-05-03 through nightly-2020-05-05
installing nightly-2020-05-05
uninstalling nightly-2020-05-05


********************************************************************************
Regression in nightly-2020-05-05
********************************************************************************
```

As you can see, 2020-05-05 is downloaded and tested thrice. Is that necessary?

spastorino:
I've noticed this and it's one of the things that should be fixed :).

chrissimpkins:
I always run with start and end dates and have not seen many of these recent issues with end only runs.  We need to look into that area of the source.

chrissimpkins:
I suggest that we look into the skip intervals when we investigate this.  I sense that we may be adding far too many steps to find the first non-fail date before the bisection starts in an average case situation.  I wonder if it is faster to initially step back by, say, one month and begin there?

hellow554:
Funny enough this still isn't fixed as of today:

```
installing nightly-2020-12-08
std for x86_64-unknown-linux-gnu: 24.58 MB / 24.58 MB [================================================================================] 100.00 % 5.77 MB/s testing...
RESULT: nightly-2020-12-08, ===> Yes
uninstalling nightly-2020-12-08

installing nightly-2020-12-01
std for x86_64-unknown-linux-gnu: 24.52 MB / 24.52 MB [================================================================================] 100.00 % 5.49 MB/s testing...
RESULT: nightly-2020-12-01, ===> No
uninstalling nightly-2020-12-01

[...]

PR #178, opened 2022-07-28
Report an estimate of remaining steps.

### LINKED ISSUE #143: feature: report expected number of steps remaining
When you run `git bisect`, it includes output like:
```
Bisecting: 337 revisions left to test after this (roughly 9 steps)
```

It would be cool if `cargo-bisect-rustc` would report the number of nightlies left to test (and expected number of steps there) and then the number of commits left to test for that nightly (and the expected number of steps there). That way I might at least quickly do back-of-envelope guess as to when I should come back and look at the result it came up with.

ehuss:
This adds some reporting to let the user know more about what cargo-bisect-rustc is doing, and provides an estimate of how many more steps it is going to do.

This is not perfect by any means, but seems to provide a decent enough approximation. Some limitations:

* It doesn't consider toolchains that fail to install (unknown_ranges). 
* It doesn't differentiate between nightlies and commits. So if you start with nightlies it will tell you N steps left, and then when it starts testing commits, the estimate starts over again.
* It doesn't include the initial start/end check in the estimate. It might be nice, but not trivial to handle.
* It doesn't consider the last check in `print_results`. Ideally that step should be removed (see #85).

Fixes #143

### DIFF
--- a/src/least_satisfying.rs
+++ b/src/least_satisfying.rs
@@ -4,10 +4,14 @@ use std::fmt;
 pub fn least_satisfying<T, P>(slice: &[T], mut predicate: P) -> usize
 where
     T: fmt::Display + fmt::Debug,
-    P: FnMut(&T) -> Satisfies,
+    P: FnMut(&T, usize, usize) -> Satisfies,
 {
     let mut cache = BTreeMap::new();
-    let mut predicate = |idx: usize| *cache.entry(idx).or_insert_with(|| predicate(&slice[idx]));
+    let mut predicate = |idx: usize, remaining, estimate| {
+        *cache
+            .entry(idx)
+            .or_insert_with(|| predicate(&slice[idx], remaining, estimate))
+    };
     let mut unknown_ranges: Vec<(usize, usize)> = Vec::new();
     // presume that the slice starts with a no
     // this should be tested before call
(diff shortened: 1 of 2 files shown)


oli-obk:
The behaviour looks good to me. The only impl concern I have is that it distracts from the actual logic. Maybe move the two new fields into a struct and add named methods for all the cases. Then also move `cache` into the struct and `predicate` to be a method on the struct.

ehuss:
I'm having a bit of difficulty fully understanding your suggestion. `remaining` and `estimate` need to be recomputed during each loop, so I'm not sure if they benefit from being in a struct.

Can you say more about what you are thinking? 

I pushed a change to move the computation into the closure so that it does not distract from the central algorithm. I also removed the hack adjusting the estimate during an `Unknown` scan, since that estimate will jump around or otherwise be wrong since `unknown_ranges` is not considered.

Otherwise I'm having a hard time seeing a way to simplify it more.

oli-obk:
Ah 🤦 I misread the logic. The new thing is much easier to grok, so mission accomplished!

oli-obk:
Ah 🤦 I misread the logic. The new thing is much easier to grok, so mission accomplished!

PR #206, opened 2022-09-01
Fix the CLI requiring the help argument

Alexendoo:
Currently the CLI on master requires the help flag to be passed:

```
$ cargo bisect-rustc
error: The following required argument was not provided: help

USAGE:
    cargo <SUBCOMMAND>

For more information try help
```
```
$ cargo-bisect-rustc
error: The following required argument was not provided: help

USAGE:
    cargo bisect-rustc [OPTIONS] [-- <COMMAND_ARGS>...]

For more information try --help
```

I copied what https://github.com/clap-rs/clap/blob/master/examples/cargo-example.rs does, with this the help message of `cargo-bisect-rustc --help` is slightly different:

```
cargo-bisect-rustc 

USAGE:
    cargo <SUBCOMMAND>

OPTIONS:
    -h, --help    Print help information

SUBCOMMANDS:
    bisect-rustc    Bisects rustc toolchains with rustup
    help            Print this message or the help of the given subcommand(s)
```

But `cargo bisect-rustc --help` works as normal

### DIFF
--- a/tests/cmd/help.toml
+++ b/tests/cmd/help.toml
@@ -1,2 +1,2 @@
 bin.name = "cargo-bisect-rustc"
-args = "--help"
+args = "bisect-rustc --help"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,4 +31,4 @@ jobs:
     - name: Verify that binary works
       run: |
         cargo run -- bisect-rustc --help | grep "EXAMPLES:"
-        cargo run -- --help | grep "EXAMPLES:"
+        cargo run -- --help | grep "SUBCOMMANDS:"
--- a/src/main.rs
+++ b/src/main.rs
@@ -57,11 +57,7 @@ const REPORT_HEADER: &str = "\
 ==================================================================================";
 
 #[derive(Debug, Parser)]
-#[clap(
-    bin_name = "cargo",
-    disable_help_flag = true,
-    subcommand_required = true
-)]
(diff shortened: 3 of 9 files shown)


tests/cmd/help.toml
```diff
@@ -1,2 +1,2 @@
 bin.name = "cargo-bisect-rustc"
-args = "--help"
+args = "bisect-rustc --help"
```
oli-obk, line 2: Please also add a test for the `--help` less variant

Alexendoo: Oh whoops I misread that, I can add a test that starts a bisect run if it's fine to have the tests download a toolchain

Alexendoo: Ah no need for downloading, can have a test that fails before that point

oli-obk: 😆 ah that's a great addition, tho not what I meant. I meant testing that `cargo-bisect-rustc` without `--help` actually prints the same thing as the one with `--help`

oli-obk: I... completely misunderstood the PR's main description XD I thought just using `cargo-bisect-rustc` should emit the help message ^^ thanks for clarifying

oli-obk: I... completely misunderstood the PR's main description XD I thought just using `cargo-bisect-rustc` should emit the help message ^^ thanks for clarifying

PR #223, opened 2022-10-03
Dependabot: Ignore patch level updates

LingMan:
Keeping them constantly updated has limited benefits while it annoys the maintainers. In the ~2.5 months Dependabot has been enabled it filed 44 PRs. With this configuration, it would have only filed 4 PRs.


oli-obk:
Thanks <3

PR #229, opened 2022-10-12
cli: Remove duplicated output of RegressOn variant names from the help

LingMan:
This was missed in the recent clap upgrade.


ehuss (approved):
Thanks!

Issue #226, opened 2022-10-07
h.toml, help.toml tests fail when HOST triple is too long


tjni:
When updating `cargo-bisect-rustc` to v0.6.4 in [nixpkgs](https://github.com/NixOS/nixpkgs/), two tests fail on aarch64-linux:

- tests/cmd/h.toml
- tests/cmd/help.toml

for the same reason:

```
21      -         --host <HOST>               Host triple for the compiler [default: [..]]
21 +         --host <HOST>               Host triple for the compiler [default:
22 +                                     aarch64-unknown-linux-gnu]
```

Because the default value for `--host` is too long, the help output wraps and no longer matches using `trycmd`'s `[..]` matcher.

Unfortunately, I'm not sure what the most elegant way is to solve this.

PR #230, opened 2022-10-12
tests: Fix tests if the HOST triple is too long

LingMan:
The HOST triple defaults to the current host and gets printed in the help output. The CLI tests currently match it with `[..]`. However, if the triple is too long clap might introduce a line break, which `[..]` can't handle.
Use `...` instead to match arbitrary lines at the position. Not elegant, but these are just smoke tests to ensure there's no major probelm. They are not meant to ensure 100% stable output.

Fixes: https://github.com/rust-lang/cargo-bisect-rustc/issues/226


LingMan:
@tjni: Could you please test this? This should fix the broken tests you were seeing, but I don't have an aarch64 system available to actually test.

tjni:
Confirmed that this fixes the issue I was seeing. Much thanks for the fast turnaround on the fix.

PR #235, opened 2022-11-01
Loosen `rust-lang/rust` remote check

clubby789:
This permits remote URLs such as `git@github.com:rust-lang/rust.git`


oli-obk:
Thanks!

PR #236, opened 2022-11-02
Try running script from working directory

### LINKED ISSUE #234: panic trying to run script, when script is given with just a filename
`cargo-bisect-rustc --script timeout.sh`

panics after the first download:

```
checking nightly-2022-10-26 from the currently installed default nightly toolchain as the last failure
checking the start range to find a passing nightly
installing nightly-2022-10-26
testing...
thread 'main' panicked at 'thiserror::Errored to run "timeout.sh": Os { code: 2, kind: NotFound, message: "No such file or directory" }', /Users/jruderman/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-bisect-rustc-0.6.4/src/toolchains.rs:302:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It should either work, or give an error message earlier saying I need to add the `./` prefix.

(The name and contents of the script don't matter for this bug; an empty script or one with just a shebang line is enough to trigger it.)

### Contrast: actually missing script

`cargo-bisect-rustc --script notfound.sh`

gives a reasonable, early error message:

`error: Invalid value "notfound.sh" for '--script <SCRIPT>': No such file or directory (os error 2)`

### Contrast: prefix script name with `./`

`cargo-bisect-rustc --script ./timeout.sh`

works.

clubby789:
Fixes #234 
If we have a non-absolute path that doesn't begin with `./`, check for the script in the current directory and adjust the script path if it exists.

### DIFF
--- a/src/toolchains.rs
+++ b/src/toolchains.rs
@@ -238,17 +238,11 @@ impl Toolchain {
                     .join(&format!("target-{}", self.rustup_name())),
             );
         }
-        let mut script = cfg.args.script.clone();
-        if let Some(path) = &script {
-            if !path.is_absolute() && !path.starts_with("./") && !path.starts_with(".\\") {
-                if let Ok(mut dir) = std::env::current_dir() {
-                    dir.push(path);
-                    if dir.is_file() {
-                        script = Some(dir)
-                    }
-                }
-            }
-        }
+        let script = cfg
+            .args
+            .script
+            .as_ref()
+            .map(|script| std::env::current_dir().unwrap().join(script));
 
         let mut cmd = match (script, cfg.args.timeout) {
             (Some(script), None) => {
--- a/src/toolchains.rs
+++ b/src/toolchains.rs
@@ -238,8 +238,19 @@ impl Toolchain {
                     .join(&format!("target-{}", self.rustup_name())),
             );
         }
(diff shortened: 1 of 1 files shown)


src/toolchains.rs
```diff
@@ -238,8 +238,19 @@ impl Toolchain {
...
+        if let Some(path) = &script {
+            if !path.is_absolute() && !path.starts_with("./") && !path.starts_with(".\\") {
+                if let Ok(mut dir) = std::env::current_dir() {
+                    dir.push(path);
+                    if dir.is_file() {
+                        script = Some(dir)
+                    }
+                }
+            }
+        }
```
ehuss, line 251: I think this could be simplified to something like:

```rust
let script = cfg.args.script.as_ref().map(|script| {
    std::env::current_dir().unwrap().join(script)
});
```

If the path is absolute, it will replace the current dir. Otherwise, it will be joined with it.

ehuss (commented):
Thanks for the PR!  I'd love to get this fixed, as I often find it awkward to need `./`.

ehuss (approved):
Thanks!

I suppose this will break if someone passes an executable from PATH, but I imagine 99% of the time that isn't the case. If that's an issue, we can revert or adjust as needed.

Issue #233, opened 2022-10-24
Wrong or incomplete error message when using non-bors git commit as --start


jruderman:
```
$ cargo-bisect-rustc --preserve --start=e67f09ac799e4a6e4e90147d62c94c9ddcc94197 --end=7aa361390867181d225a7775340f50e8162e16a2
bisecting ci builds
starting at e67f09ac799e4a6e4e90147d62c94c9ddcc94197, ending at 7aa361390867181d225a7775340f50e8162e16a2
opening existing repository at "rust.git"
Found origin remote under name `origin`
refreshing repository at "rust.git"
remote: Enumerating objects: 209, done.
remote: Counting objects: 100% (180/180), done.
remote: Compressing objects: 100% (53/53), done.
remote: Total 209 (delta 148), reused 140 (delta 127), pack-reused 29
Receiving objects: 100% (209/209), 165.35 KiB | 2.71 MiB/s, done.
Resolving deltas: 100% (149/149), completed with 88 local objects.
From https://github.com/rust-lang/rust
   7aa36139086..7feb003882e  master     -> origin/master
fetching (via local git) commits from e67f09ac799e4a6e4e90147d62c94c9ddcc94197 to 7aa361390867181d225a7775340f50e8162e16a2
opening existing repository at "rust.git"
Found origin remote under name `origin`
refreshing repository at "rust.git"
looking up first commit
looking up second commit
checking that commits are by bors and thus have ci artifacts...
ERROR: failed during attempt to create/access local git repository
```

It should say what problem it caused the bisector to exit: a non-bors commit was specified as start.

(Or it could guess some appropriate bors commit to use in place of the one I specified.)

### Version

```
cargo-bisect-rustc --version
cargo-bisect-rustc 0.6.4
```

on mac

ngc0202:
I just spent too long debugging the same exact issue. The issue is in `main` they're `Display` printing the error, which shows the minimal amount of info instead of pretty-printing or `Debug` printing which shows the complete cause.

ehuss:
I think that between #103 and #237, this is resolved.

However, there is an issue with github access that causes it to panic. I have opened #318 to fix that.

PR #237, opened 2022-11-07
Use debug printing for anyhow errors to show the full cause and context

ngc0202:
The top-level anyhow context is often not enough for the user to fully diagnose the issue that's occurring, such as in issue #233. This changes from `Display` to `Debug` printing to show the full context and backtrace if available.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1165,7 +1165,7 @@ fn main() {
             Ok(ExitError(code)) => process::exit(code),
             Err(err) => {
                 let error_str = "ERROR:".red().bold();
-                eprintln!("{} {}", error_str, err);
+                eprintln!("{} {:?}", error_str, err);
                 process::exit(1);
             }
         }


ngc0202:
As an example of the difference this makes, here's the before and after stderr output from when the commit isn't from bors as is expected:

Before:
```
ERROR: failed during attempt to create/access local git repository
```

After:
```
ERROR: failed during attempt to create/access local git repository

Caused by:
    Expected author Brian Anderson to be bors for beb9a0dfc52ebda4f8db4e5d439e08e4f3a43a39.
     Make sure specified commits are on the master branch!
```

ngc0202:
On a side not related to the error I posted as an example, it seems like this merge request was accidentally merged by someone other than `bors` which breaks the assumption that this tool makes about merges, meaning you seemingly can't bisect across this boundary. Should there by an exception made in the code here?

oli-obk:
We should just ignore those merge commits instead of erroring on them I think.

PR #243, opened 2022-12-03
Fix bounds checking with missing bound.

tjni:
Before this, checking that start and end are both before the current date when either one of start or end is missing didn't work.

This was detected by the command-line tests that are run when building this in [nixpkgs](https://github.com/NixOS/nixpkgs).


ehuss (approved):
Thanks!

PR #251, opened 2023-01-21
Switch the action for installing rust

ehuss:
The actions-rs/toolchain action is no longer maintained and is generating several deprecation warnings. This switches to dtolnay/rust-toolchain, which should be more reliable.


LingMan (approved):
That's one thing off my todo list :D

PR #252, opened 2023-01-21
Merge fmt workflow with the main CI workflow

ehuss:
This just makes it a little easier to navigate the GitHub UI so that it shows all of the jobs in one "run" instead of having separate runs.


.github/workflows/ci.yml
```diff
@@ -32,3 +32,15 @@ jobs:
...
+
+  fmt:
+    name: rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Run rustfmt check
+        run: |
+          rustup component add rustfmt
```
LingMan, line 44: Maybe use `dtolnay/rust-toolchain` here as well so it always uses stable rust?
```
    - name: Install rustfmt
      uses: dtolnay/rust-toolchain@stable
      with:
        components: rustfmt
```

.github/workflows/ci.yml
```diff
@@ -32,3 +32,15 @@ jobs:
...
+          cargo fmt --version
+          cargo fmt -- --check || (echo "Please reformat your code with 'cargo fmt' (version $(cargo fmt --version))"; false)
```
LingMan, line 46: Since rustfmt has grown its own direct check flag since this line was originally written:
```suggestion
          cargo fmt --check || (echo "Please reformat your code with 'cargo fmt' (version $(cargo fmt --version))"; false)
```

.github/workflows/ci.yml
ehuss: Sounds reasonable. I went ahead and merged the changes of #251 with this.

LingMan: Hm, my intention was to replace the raw `rustup` command with `dtolnay/rust-toolchain`'s `components` parameter. Although it probably does just call rustup internally so at this point it's mostly a question of cosmetics.

ehuss: Oh, I misunderstood. Yea, just using `components` seems nicer.

LingMan (approved):

PR #262, opened 2023-02-27
Add options to skip verifying bisection ranges

### LINKED ISSUE #258: feature: Add option to skip verifying the start/end of a range
It would be useful to  skip verification bisection start/end if (using Playground or otherwise) we already know a valid regression range.

clubby789:
Closes #258

Adds two options for bypassing verifying the start and end of nightly and CI ranges. If no start nightly has been provided, the start date will be detected as usual.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -179,6 +179,18 @@ a date (YYYY-MM-DD), git tag name (e.g. 1.58.0) or git commit SHA."
 
     #[arg(long, help = "Do not install cargo [default: install cargo]")]
     without_cargo: bool,
+
+    #[arg(
+        long,
+        help = "Do not verify that the start and end of the nightly range are correct"
+    )]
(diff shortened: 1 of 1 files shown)


clubby789:
Yeah, if you know for sure that a regression happened between (for example) stable and beta, you can save some time by skipping checking those

oli-obk:
Hmm... that needs a bit more documentation on the flags. To me it sounded like verifying that the commit/nightly exists or somehow is valid in itself. But this is solely about not running the bisection on the start/end and only running it between the given values

Also: why do we have different flags for CI artifacts and nightly?

clubby789:
I'll update the help to document this better. I made two flags so that we don't skip verifying the CI range if the user has only provided nightly dates. It would probably make more sense to have it be a single flag that skips verifying whichever of the two the user has provided.

PR #266, opened 2023-03-20
Add a user guide with examples

### LINKED ISSUE #264: Structured documentation and examples
@oli-obk Would you be ok with me moving the cargo-bisect-rustc documentation into an mdbook published on GitHub Pages? I'd like to organize the documentation better. And more importantly I want to collect a variety of recipes (examples) of how to do various things. For example, I often forget how to do more complex tasks on Windows, and having something I could just copy/paste would save a lot of time.

ehuss:
This replaces the old tutorial with a new book that goes into more detail on various topics. It also contains a suite of examples to show different uses.

Closes #264


ehuss:
There is a preview of what this looks like at: https://ehuss.github.io/cargo-bisect-rustc/introduction.html

.github/workflows/docs.yml
LingMan, line 1: Github has published starter workflows for publishing with Github Actions. I'd probably base this one on the mdBook workflow they've published instead of inventing my own.

Docs: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow
mDbook starter workflow: https://github.com/actions/starter-workflows/blob/main/pages/mdbook.yml

I also suspect that pinning the mdBook version will quickly result in outdated versions being used to build. Pretty sure Dependabot won't open PRs to update it.
One option would be `cargo install mdbook` but running that on each push is a waste of cpu cycles. Instead you could install it with `taiki-e/install-action`. That action gets regular updates so far, so I'd trust it more to stay up-to-date until proven otherwise.

Usage would be:
```
- uses: taiki-e/install-action@v2
  with:
    tool: mdbook
```

ehuss:
@oli-obk Just checking if you had any thoughts on this?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,38 @@
...
+      contents: write  # for Git to git push
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Install mdbook
+      run: |
+        mkdir mdbook
+        curl -Lf https://github.com/rust-lang/mdBook/releases/download/v0.4.28/mdbook-v0.4.28-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
```
oli-obk, line 22: maybe put the version into an env var so it is easy to update?

oli-obk (changes requested):
Looks good to me. I didn't check that everything was moved, but the summary looks good

oli-obk (approved):

PR #280, opened 2023-07-03
Fix bisecting into rollups via unrolled perf builds

lqd:
The perfbot comment on rollups has [recently changed format](https://github.com/rust-lang/rustc-perf/pull/1623), breaking the commit extraction from that comment.

Also updates proc-macro2 to fix build errors on nightly.

Now uses a regex to look for the shas found in the perf table column, to support both formats:
- https://github.com/rust-lang/rust/pull/113014#issuecomment-1605868471
- https://github.com/rust-lang/rust/pull/113105#issuecomment-1610393473

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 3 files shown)


lqd:
Converting to draft to make more changes: the parsing here needs to be backwards compatible with the prior format.

lqd:
I've added the 2 linked examples as tests.
Should be ready now.

PR #286, opened 2023-07-26
switch CI downloads to use CI artifacts URL

lqd:
If I understand correctly, https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rustc.20nightly.20download.20speed/near/378145048 means that `cargo-bisect-rustc` should be using the CI artifacts URL to allow for better caching and CDNs.

draft until confirmation from @pietroalbini that this is needed, and @bjorn3 to see if this fixes their S3 download speed


lqd:
@bjorn3's download speeds would be improved by this PR according to https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rustc.20nightly.20download.20speed/near/378786646

pietroalbini (approved):

PR #295, opened 2023-10-13
Use rustls instead of openssl

Nilstrieb:


LingMan:
Are there really no changes to `Cargo.lock`?

Nilstrieb:
godd point, I should run this locally. Looks like CI doesn't use locked

Nilstrieb:
cargo-bisect-rustc uses git2, which uses openssl through libgit2. We could probably move to gitoxide at some point but I don't want to do that right now.

PR #307, opened 2024-01-23
Default to `--access github`

Nilstrieb:
I've been using `cargo-bisect-rustc` for a while now, and `--access
github` has never caused any problems. Maybe it can run into
ratelimiting in extreme cases, but the user should then just use `git`.

Meanwhile, git is a very confusing default as it will suddelny just
start cloning the repo unless you were very careful when installing.

### DIFF
--- a/guide/src/rust-src-repo.md
+++ b/guide/src/rust-src-repo.md
@@ -3,10 +3,30 @@
 For `cargo-bisect-rustc` to work, it needs to be able to read the git log of the [`rust-lang/rust`] repo.
 `cargo-bisect-rustc` supports several methods for this described below.
 
+## GitHub API
+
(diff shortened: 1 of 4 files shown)


ehuss:
I have not used the github access method much, but I'm game to try. It certainly will be a better default experience as long as it is not rate limited. @oli-obk what do you think?

Nilstrieb:
I think it is rate limited, but I've never hit the limit in practice when doing one-off bisections (or even two or three).

Nemo157:
The unauthenticated rate limit is 60 requests/hour/IP, which should be more than enough for bisecting, outside heavily NATted connections (I would assume it only does one request per run?).

PR #311, opened 2024-02-06
Improve debug wording

apiraino:
A user recently mentioned that the wording of cargo bisect during the bisection is a bit unclear (https://togithub.com/rust-lang/rust/issues/119822#issuecomment-1929153702), namely what do the messages after every test mean.

With this patch I'm trying to make it clearer (I hope I got the logic right).

thanks for checking this out!

r? @ehuss

### DIFF
--- a/src/least_satisfying.rs
+++ b/src/least_satisfying.rs
@@ -174,7 +174,11 @@ pub enum Satisfies {
 
 impl fmt::Display for Satisfies {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            Self::Yes => write!(f, "Reproduces the regression"),
+            Self::No => write!(f, "Does not reproduce the regression"),
+            Self::Unknown => writeln!(f, "Unable to figure out if it reproduces the regression"),
+        }
     }
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -1159,7 +1159,9 @@ impl Config {
 
         if !toolchains.is_empty() {
             // validate commit at start of range
-            eprintln!("checking the start range to verify it passes");
+            eprintln!(
+                "checking the start range: testing the sample code, should not reproduce the regression"
+            );
             let start_range_result = self.install_and_test(&toolchains[0], &dl_spec)?;
             if start_range_result == Satisfies::Yes {
                 bail!(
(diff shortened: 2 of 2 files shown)


src/least_satisfying.rs
```diff
@@ -174,7 +174,11 @@ pub enum Satisfies {
 
 impl fmt::Display for Satisfies {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            Self::Yes => write!(f, "Does not reproduce the regression"),
+            Self::No => write!(f, "Reproduces the regression"),
```
ehuss, line 179: I think this might be backwards.

`Yes` means it reproduces the regression (rustc failed to compile).
`No` means it did not reproduce (rustc compiled successfully).

However, the terms "regress" and "not regress" may not be completely clear. For example, you might be looking for when something was *fixed*, such as when using `--regress=no-error`, which has an opposite meaning ("no" means "fails to compile", "yes" means "compiles successfully").

I would be fine with merging this with the text reversed, since it is a little clearer than "yes" and "no", but I suspect there is still room for confusion here. I think longer term it would be better to have customizable terms, and then things like `--regress=no-error` could change them to "compiles successfully" and "failed to compile", and offer the ability to customize the message like `git bisect` does (maybe `--term-old=good --term-new=bad`).

apiraino: thanks @ehuss exactly the feedback I needed 

Ok I've fixed the "logical" error.

Now, for a better wording. The Yes/No in the end means whether the toolchain at hand did produce the expected result using the sample code provided, correct? "Yes" (it did) or "No" (it didn't).

I couldn't think of a single sentence to squeeze all "success" cases and all "failure" cases into a single sentence. I also think that the message should be contextual to the `--regress` parameter used.

Unless we use some vague wording like "_requested condition matched_" or "_not matched_".

wdyt ?

ehuss: > Unless we use some vague wording like "requested condition matched" or "not matched".

That actually sounds a little better to me. I do worry that specifically saying `Reproduces the regression` could be confusing if you are searching for when the regression was fixed (since it will be backwards).

I opened #316 for being able to customize it. I can work on that sometime later unless you are interested.

apiraino:
This was handled (better) in #337 and #339

PR #323, opened 2024-03-02
Do not unconditionally set the target.

### LINKED ISSUE #284: can cargo-bisect-rustc respect the .cargo/config.toml?
can cargo-bisect-rustc respect the .cargo/config.toml?  (Should it already be doing so?)

As part of dissecting https://github.com/rust-lang/rust/issues/111888, I was looking into adding a linker flag to the rustc invocations via a `.cargo/config.toml` file, as described here: https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

But it seems from my local experiments that while my manual invocations of `cargo` are pulling in the config settings, the invocations that `cargo-bisect-rustc` emits do not seem to be including the rustflags specified there.

ehuss:
This removes the `CARGO_BUILD_TARGET` environment variable which causes cargo to unconditionally work in "cross compile mode". This is different from a normal `cargo` invocation without the `--target` flag, in which case it runs in "host mode", where cargo and other tools have distinctly different behavior. This can be surprising when you get one behavior with `cargo build` and a different one with `cargo bisect-rustc`.

This was added in #159 to address some issue with using `cross-rs` with `cargo bisect-rustc`. I'm not sure why, as I would think the recommended behavior when using cross is to set the `--target` flag. The given PR didn't really explain the situation.

Closes #284

### DIFF
--- a/guide/src/examples/rustdoc.md
+++ b/guide/src/examples/rustdoc.md
@@ -21,7 +21,7 @@ set -e
 
 cargo doc
 
-grep "some example text" $CARGO_TARGET_DIR/$CARGO_BUILD_TARGET/doc/mycrate/fn.foo.html
+grep "some example text" $CARGO_TARGET_DIR/doc/mycrate/fn.foo.html
 ```
 
 This can be used with the `--script` option:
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum (approved):

ehuss:
I realize I made a mistake, it should set the target if the user sets `--target` on the command-line.

PR #325, opened 2024-03-05
Refactor bounds definition.

### LINKED ISSUE #247: Automatic end-date calculation fails during promotion window
If the `--end` date is not specified, then cargo-bisect-rustc will [assume the end date is today](https://github.com/rust-lang/cargo-bisect-rustc/blob/1b45158bd499c942f83fb9fd326cacd66933ace1/src/main.rs#L867-L868). However, there is a window of an hour or so from UTC midnight until nightly gets published that this will result in an error:

```
ERROR: Could not find nightly-2023-01-20; url: https://static.rust-lang.org/dist/2023-01-20/rustc-nightly-x86_64-apple-darwin.tar.gz
```

I think it would be good if when `--end` is not specified that cargo-bisect-rustc would walk back from "today" at least a few days to find a nightly that exists.  There is already code for it to skip over missing nightlies while bisecting. I'm not sure if there is something that can be reused, or if there are other options.

cargo-bisect-rustc 0.6.5

ehuss:
This is a refactoring of the bounds definitions in hopes to make the code easier to follow. I've struggled following it in the existing code because there are several mutations of `Opts::start` and `Opts::end`, and the logic for interpreting the bounds was scattered in several places.

A general overview of the change:
* Moved the bounds code to a separate module to try to clean things up.
* Added a `Bounds` enum to represent the start and end boundaries. All code that accessed the `start` and `end` values directly now looks at `Config::bounds`.
* All the interpretation of `--start` and `--end` happens in `Bounds::from_args` in a single match.
* Fixed a small bug where it would panic with `--by-commit` and no `--start` (it now shows an error).
* Added some UI tests for the various simple error cases.

This also fixes a bug I somehow run into several times. During the time between midnight UTC and whenever the nightlies are published, cbr will fail if you don't specify an `--end`. This updates it to fetch the manifest from the network to determine the date of the latest nightly (`find_latest_nightly`) instead of assuming the current date. This requires pulling in `toml` to parse the manifest.

Fixes #247

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -212,15 +212,13 @@ fn installed_nightly_or_latest() -> anyhow::Result<GitDate> {
 
 /// Returns the date of the latest nightly (fetched from the network).
 fn find_latest_nightly() -> anyhow::Result<GitDate> {
-    let url = format!("{NIGHTLY_SERVER}/channel-rust-nightly.toml");
+    let url = format!("{NIGHTLY_SERVER}/channel-rust-nightly-date.txt");
     eprintln!("fetching {url}");
     let client = Client::new();
-    let mut response = download_progress(&client, "nightly manifest", &url)?;
-    let mut manifest = String::new();
-    response.read_to_string(&mut manifest)?;
-    let manifest: toml::Value = toml::from_str(&manifest)?;
-    let date = manifest["date"].as_str().expect("date is a string");
-    let date = NaiveDate::parse_from_str(date, "%Y-%m-%d")?;
+    let mut response = download_progress(&client, "nightly date", &url)?;
+    let mut body = String::new();
+    response.read_to_string(&mut body)?;
+    let date = NaiveDate::parse_from_str(&body, "%Y-%m-%d")?;
     eprintln!("determined the latest nightly is {date}");
     Ok(date)
 }
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -1,0 +1,224 @@
+//! Definitions of bisection bounds.
+
+use crate::toolchains::{
+    download_progress, parse_to_naive_date, Toolchain, NIGHTLY_SERVER, YYYY_MM_DD,
+};
+use crate::GitDate;
+use crate::Opts;
+use crate::{today, EPOCH_COMMIT};
+use anyhow::bail;
+use chrono::NaiveDate;
+use reqwest::blocking::Client;
+use std::io::Read;
+use std::str::FromStr;
+
+/// A bisection boundary.
+#[derive(Clone, Debug)]
+pub enum Bound {
+    Commit(String),
+    Date(GitDate),
+}
+
+impl FromStr for Bound {
+    type Err = std::convert::Infallible;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        parse_to_naive_date(s)
+            .map(Self::Date)
+            .or_else(|_| Ok(Self::Commit(s.to_string())))
+    }
+}
+
+impl Bound {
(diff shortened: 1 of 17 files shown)


ehuss:
@Mark-Simulacrum Is this something you would be able to review?

src/bounds.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+/// The starting bisection bounds.
+pub enum Bounds {
+    /// Indicates to search backwards from the given date to find the start
+    /// date where the regression does not occur.
+    SearchNightlyBackwards { end: GitDate },
+    /// Search between two commits.
+    Commits { start: String, end: String },
```
Mark-Simulacrum, line 64: IIUC, these aren't actually commits -- or at least indirectly so. E.g., `origin/master` is valid. Maybe we can note that in the comment?

src/bounds.rs
```diff
@@ -0,0 +1,224 @@
...
+        // even try to convert the other bound to a date.
+        return Ok((args.start.clone(), args.end.clone()));
+    }
+    let fixup = |which: &str, bound: &Option<Bound>| -> anyhow::Result<Option<Bound>> {
+        if is_tag(bound) {
+            if let Some(Bound::Commit(tag)) = bound {
+                let date = args
+                    .access
+                    .repo()
+                    .bound_to_date(Bound::Commit(tag.clone()))?;
```
Mark-Simulacrum, line 171: This feels wrong to me. If we're looking up (say) 1.68.0, we presumably need something here to say that we want the *stable* release as of that date or otherwise go from the stable release to the nightly date range it "points" at?

Otherwise we're missing ~6 weeks here, right?

src/bounds.rs
```diff
@@ -0,0 +1,224 @@
...
+fn installed_nightly_or_latest() -> anyhow::Result<GitDate> {
+    if let Some(date) = Toolchain::default_nightly() {
+        return Ok(date);
+    }
+    find_latest_nightly()
+}
+
+/// Returns the date of the latest nightly (fetched from the network).
+fn find_latest_nightly() -> anyhow::Result<GitDate> {
+    let url = format!("{NIGHTLY_SERVER}/channel-rust-nightly.toml");
```
Mark-Simulacrum, line 213: https://static.rust-lang.org/dist/channel-rust-nightly-date.txt maybe works? Not sure if it's uploaded late enough in the process...

Mark-Simulacrum (changes requested):

src/bounds.rs
ehuss: The `RustRepositoryAccessor::commit` function returns the **merge base** with the master branch. So it essentially looks for the point where the given tag branched off the master branch. It does not return the commit associated with that tag.

That means it does not include any fixes merged to the `beta` branch, but should provide a close approximation of what the equivalent nightly was.

(This particular code is unchanged from the original, and has been working sufficiently well for me.)

src/bounds.rs
ehuss: Oh, that's better! I had forgotten about that.

From what I can tell, it looks like it should work, so I went ahead and switched to the txt file. From what I can see in promote-release, the [last step](https://github.com/rust-lang/promote-release/blob/dd7a74323e1a3b489d452620732574098afefc76/src/main.rs#L248) it takes involves copying the manifests, including this txt file, and that happens after the archives have been uploaded.

src/bounds.rs
Mark-Simulacrum: Ah, I missed that aspect. Sounds good to me.

Mark-Simulacrum (approved):

PR #334, opened 2024-04-16
Stop stripping and LTO

Nilstrieb:
**Stop stripping**:
Stripping symbols here is bad, because it makes the binary completely opaque and undebbuggable, should something ever go wrong. The binary size benefit is very minor, this program is supposed to run on desktop computers with tons of disk space. The tradeoff is bad.

I like to follow the rule of "do not ever strip symbols unless you absolutely need to", at least for programs on dev systems.

**Stop LTO**:
The same binary size non-concerns as stripping apply. Also, speed basically doesn't matter here. Meanwhile, we want this program to be `cargo install`-able, so compile times matter. Again, the tradeoff is bad.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,6 @@ repository = "https://github.com/rust-lang/cargo-bisect-rustc"
 version = "0.6.8"
 edition = "2021"
 
-[profile.release]
-strip = "symbols"
-lto = "fat"
-
 [dependencies]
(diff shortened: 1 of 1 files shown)


Nilstrieb:
If anyone cares about sacrificing things for 4 megabytes like in #157 they can always manually override the setting for their builds, this is just the default.

Nilstrieb:
@oli-obk you have approved the other pr

PR #335, opened 2024-04-25
Add a flag for pretending to be a stable compiler when bisecting.

oli-obk:
This should avoid having to patch ahash and similar crates to not do nightly feature detection and break the bisection (see https://github.com/rust-lang/rust/issues/123276#issuecomment-2075001510 for an example).

We could also do this by default and add an opt-out flag for when you want to bisect code that uses feature gates

### DIFF
--- a/src/toolchains.rs
+++ b/src/toolchains.rs
@@ -257,12 +257,7 @@ impl Toolchain {
             }
             (None, None) => {
                 let mut cmd = Command::new("cargo");
-                cmd.arg(&format!("+{}", self.rustup_name()));
-                if cfg.args.command_args.is_empty() {
-                    cmd.arg("build");
-                } else {
-                    cmd.args(&cfg.args.command_args);
-                }
+                self.set_cargo_args_and_envs(&mut cmd, cfg);
                 cmd
             }
             (Some(script), Some(timeout)) => {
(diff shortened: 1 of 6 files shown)


apiraino:
thank youuu :pray:

src/toolchains.rs
```diff
@@ -322,6 +312,32 @@ impl Toolchain {
...
+            cmd.arg("build");
+        } else {
+            cmd.args(&cfg.args.command_args);
+        }
+        if cfg.args.pretend_to_be_stable && self.is_current_nightly() {
+            // Forbid using features
+            cmd.env(
+                "RUSTFLAGS",
+                format!(
+                    "{} -Zenable-features=",
```
ehuss, line 328: Shouldn't this be `allow-features`?

src/toolchains.rs
```diff
@@ -322,6 +312,32 @@ impl Toolchain {
...
+                    std::env::var("RUSTFLAGS").unwrap_or_default()
+                ),
+            );
+            // Make rustc report a stable version string derived from the current nightly's version string.
+            let version = rustc_version::version_meta().unwrap().semver;
+            cmd.env(
+                "RUSTC_FORCE_RUSTC_VERSION",
+                format!("{}.{}.{}", version.major, version.minor, version.patch),
```
ehuss, line 336: This doesn't actually change the version string, does it? My understanding is that it only changes the stable crate id and incremental cache version. I'm not sure what benefit that brings.

oli-obk: I tested it on the rustc side, and it works: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0bf99dd363de2b1659c24e9e357a34e7

oli-obk: hmm looks like I screwed up testing

lochetti:
Hey! Sorry to ask, but anyone know if we could merge/continue with this PR?  :) 

Thanks!

ehuss:
@oli-obk Is this something you want to continue with? I forget what the status is.

PR #337, opened 2024-05-22
Improve terms defaults

apiraino:
hey @ehuss :wave: 

this patch tries to implement your suggestion (see [comment](https://github.com/rust-lang/cargo-bisect-rustc/pull/330#discussion_r1583635275)) implementing reasonable defaults for `--term-old` and `--term-new`.

I've tried doing something that would not lead to confusion but I find the implementation confusing the same :smile: 

Is it just me? Any suggestion on how to improve?

Thanks

### DIFF
--- a/src/least_satisfying.rs
+++ b/src/least_satisfying.rs
@@ -175,26 +175,49 @@ pub enum Satisfies {
 }
 
 impl Satisfies {
-    pub fn msg_with_context(&self, regress: &RegressOn, term_old: &str, term_new: &str) -> String {
-        let msg_yes = if regress == &RegressOn::Error || regress == &RegressOn::Ice {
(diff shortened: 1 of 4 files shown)


ehuss:
Hm, doing some testing here I still find several situations where this displays the opposite of what it should.

For example, when testing something that fails to compile on old compilers, and compiles successfully on newer ones, and using the `--regress=success` flag, this PR changes it so that it always prints "Compile error" for both cases.

I posted https://github.com/rust-lang/cargo-bisect-rustc/pull/339 with what my intentions were about how the old/new terms could be defaulted based on the regression mode.

ehuss:
Thanks @apiraino! I went ahead and merged #339, so I'm going to go ahead and close this. I appreciate you helping with this.

PR #354, opened 2024-11-11
Add FileCheck mode

clubby789:
This allows using, e.g. `cargo-bisect-rustc --start 1.70.0 --filecheck src/lib.rs --preserve --regress success` to investigate codegen regressions/improvements


tests/cmd/h.stdout
clubby789, line 1: I'm not really sure why this changed

apiraino: These files are autogenerated by `trycmd` when running tests, based on the content of the cli options.

I forgot the details, but in [tests/README.md](https://github.com/rust-lang/cargo-bisect-rustc/tree/master/tests) I added some info a while ago

clubby789: I meant I'm not sure specifically why the formatting of the help output changed here - descriptions are now all placed on the following line

apiraino: I did some bisection of your patchset. The change that trigger this is the new command line argument
```
    #[arg(long, help = "Path to LLVM FileCheck")]
    filecheck_path: Option<PathBuf>,
```

if you remove the underscore from the parameter name (for example to `filecheckpath`), the help  output does not add a newline anymore.

So, the issue is with `clap`. I quickly skimmed the changelog but couldn't find anything explaining this strange thing. Feel free to investigate further  :smile:

clubby789: Ah it's not the underscore, it's the name length - I think when the size of the longest arg name + the longest help string are such that having all arguments inline would be very long, it's wrapped round to a new line
Repro: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ba39c32190f89d5b28976c14b28e58cb
Remove the `S` from the argument name and it'll be in-line

PR #356, opened 2024-11-12
Print all args properly in final report

### LINKED ISSUE #254: 'Reproduce with' output has invalid command
I invoked the program with `cargo-bisect-rustc --start 2023-01-27 --regress ice`. The output of the `reproduce with` section, however, was

```bash
cargo bisect-rustc 2023-01-27 --regress ice 
```
Missing the `--start` portion of the command.

clubby789:
Fixes #254

When invoked with `cargo-bisect-rustc`, our args look like
`["cargo-bisect-rustc", "--start", ...]`
However, when invoked with `cargo bisect-rustc`, our args look like
`["/path/to/cargo-bisect-rustc", "bisect-rustc", "--start", ...]`

Skip past these initial arguments instead of hardcoding the index

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -695,10 +695,11 @@ fn print_final_report(
     eprintln!("Reproduce with:");
     eprintln!("```bash");
     eprint!("cargo bisect-rustc ");
-    for (index, arg) in env::args_os().enumerate() {
-        if index > 1 {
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Issue #359, opened 2024-12-27
Unexpected arguments are incorrectly identified


dtolnay:
Something about how the command-line arguments are being parsed with Clap makes worse errors than other Cargo subcommand usages of Clap.

This error confusingly says that `bisect-rustc` is unexpected, not that `--components` is unexpected. (The correct flag is called `--component` and Clap should know to suggest this.)

```console
$ cargo bisect-rustc --components=clippy
error: unexpected argument 'bisect-rustc' found

Usage: cargo bisect-rustc [OPTIONS] [-- <COMMAND_ARGS>...]

For more information, try '--help'.
```

Compared to these, which both correctly identify `--components` as unexpected.

```console
$ cargo expand --components=clippy
error: unexpected argument '--components' found

  tip: to pass '--components' as a value, use '-- --components'

Usage: cargo expand [OPTIONS] [ITEM]

For more information, try '--help'.
```

```console
$ cargo docs-rs --components=clippy
error: unexpected argument '--components' found

  tip: a similar argument exists: '--open'

Usage: cargo docs-rs <--open|--verbose|--color <WHEN>|--version|--package <SPEC>|--jobs <N>|--target <TARGET>|--target-dir <DIRECTORY>|--manifest-path <PATH>|--frozen|--locked|--offline>

For more information, try '--help'.
```

PR #360, opened 2024-12-28
Give --help example of using grep on stderr of rustc

Enselic:
It is very useful in cases like [this](https://github.com/rust-lang/rust/issues/93710#issuecomment-2564259274) but it is a bit tricky to figure out how to do it if you are new to using cargo-bisect-rustc.

By adding this example to --help we can make it easier for others to discover this trick.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,6 +91,13 @@ enum Cargo {
     ```
     cargo bisect-rustc --start 6a1c0637ce44aeea6c60527f4c0e7fb33f2bcd0d \\
       --end 866a713258915e6cbb212d135f751a6a8c9e1c0a --test-dir ../my_project/ --prompt -- build
+    ```
+
+    Run `rustc` directly (without involving `cargo`) and use the presence of 'Undefined' in stderr
+    as the definition for a regression:
+    ```
+    cargo bisect-rustc --start=2023-03-31 --end=2023-12-21 --script bash -- -c \\
+      \"rustc -Cpasses=lint -Cinstrument-coverage --crate-type=dylib test.rs 2>&1 | grep Undefined\"
     ```"
 )]
 #[allow(clippy::struct_excessive_bools)]
--- a/tests/cmd/h.stdout
+++ b/tests/cmd/h.stdout
@@ -48,3 +48,10 @@ Examples:
     cargo bisect-rustc --start 6a1c0637ce44aeea6c60527f4c0e7fb33f2bcd0d /
       --end 866a713258915e6cbb212d135f751a6a8c9e1c0a --test-dir ../my_project/ --prompt -- build
     ```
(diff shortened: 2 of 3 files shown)


ehuss:
I'm wondering, have you seen the documentation at https://rust-lang.github.io/cargo-bisect-rustc/? I'm a little reluctant to add lots of examples here, since there are many different workflows. What do you think of linking to that documentation in the help instead?

Enselic:
Thanks for taking a look at my PR.

Yep, I am aware of those docs. If strongly prefer a summary of examples in `--help` over being redirected to a web page with one example per page, however. So I think it's great that `--help` already has a couple of examples.

I think we should make it as easy as possible for casual users of `cargo-bisect-rustc` to use the tool. Because that means we are more likely to get help with bisecting in e.g. the context of issue triaging. The first time someone encounters the tool, I think the web documentation is probably a good entry point. But once they have used to tool a few times, they just need a brief refresher on how to use the tool. That they already have the tool installed I think is a sufficiently strong indication that they are not encountering the tool for the first time. And if they type `--help`, why whould we "hold back" information and force them to take one (or actually several, since the web docs are multi-paged) steps to see some examples? Examples are a great way to quickly demonstrate usage without a user having to read the docs for each individual option and figure out how to combine them.

And I think my example is actually a great example to have in `--help`, because it demonstrates something non-obvious but important, namely that that you can use `--script bash -- -c "..."` to unlock the use of `grep` for bisecting. **Without** having to create a separate `.sh`-script which also has the downside of requiring extra steps, which unnecessarily detracts users of the tool.

apiraino:
@Enselic I am also a great fan of offline documentation but at this time the offline docs for this tool are a bit clunky to use (see also #359). I agree that for now it's probably quicker to just update the online documentation.

But.

If your intent is to expand offline documentation, how about giving this tool a `man` page that replicates and expands the online help? I didn't investigate deeply but `clap` supports [auto-generating man pages](https://github.com/clap-rs/clap/tree/master/clap_mangen) using the `roff` crate. Or like [bat](https://github.com/sharkdp/bat/blob/18d963f4232dfde7ce4b54a75fbd2acc78f85312/build/application.rs#L5) does.

Ideally a workflow that generates both offline and online help (can I dream? :slightly_smiling_face: )

Enselic:
> intent is to expand offline documentation

Well my intent is to make the tool easier to use by providing a useful example directly under `--help`. I'm not a big fan of `man` pages because while all installations of `cargo-bisect-rustc` has `--help` (e.g. `cargo install --path .` from git or simply `cargo install cargo-bisect-rustc`), not all installations have a `man` page installed. So to me it is clear that `--help` is the best place to put short but informative examples, because it is available to all users in a convenient way.

PR #362, opened 2024-12-31
Make `cargo-bisect-rustc` work with hyphen or space

### LINKED ISSUE #359: Unexpected arguments are incorrectly identified

ehuss:
This makes it so that the executable name `cargo-bisect-rustc` (with a hyphen) works the same as `cargo bisect-rustc` (with a space). This fixes parsing problems when used with a space as demonstrated in #359.

Previously, cbr would try to support both forms, but it didn't always work (see #359). This is due to the way it was trying to do a fallback parsing without the subcommand, which means CLI parsing errors when used as a subcommand to not work (since the original error was discarded).

The solution is to not attempt to use clap to do subcommand handling, and instead just strip the `bisect-rustc` subcommand if it exists.

Other than fixing the errors, one observable change is that `cargo-bisect-rustc` (without a space, with no arguments) will now behave the same as `cargo bisect-rustc` (it will start the bisection process). Previously it would print a somewhat confusing error about the missing `bisect-rustc` subcommand. This also apples to `--help` and `-h`.

Fixes #359

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -70,12 +70,6 @@ const REPORT_HEADER: &str = "\
 ==================================================================================";
 
 #[derive(Debug, Parser)]
-#[command(bin_name = "cargo", subcommand_required = true)]
-enum Cargo {
-    BisectRustc(Opts),
-}
-
-#[derive(Debug, Parser)]
 #[command(
     bin_name = "cargo bisect-rustc",
     version,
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -397,16 +391,15 @@ impl Config {
...
-            }
-            _ => Opts::parse(),
-        },
-    };
+    let mut os_args: Vec<_> = std::env::args_os().collect();
+    // This allows both `cargo-bisect-rustc` (with a hyphen) and
+    // `cargo bisect-rustc` (with a space) to work identically.
+    if let Some(command) = os_args.get(1) {
+        if command == "bisect-rustc" {
+            os_args.remove(1);
```
oli-obk, line 399: Weird that we don't need to adjust arg 0 to show "the right" (whatever that means) command. The output seems to look right

oli-obk (approved):

PR #374, opened 2025-04-01
Add another example for hanging compilation

apiraino:
While bisecting [rust#139197](https://togithub.com/rust-lang/rust/issues/139197), I spent more time than I wanted to figure out how to bisect a hanging compilation of some code that should not compile anyway.

Could this case be useful in the documentation?

Thanks


ehuss (approved):
Thanks!

PR #381, opened 2025-05-20
Use `rust-lang/rust` instead of `rust-lang-ci/rust` for unrolled rollup build commit URLs

Kobzol:
We have moved CI from `rust-lang-ci/rust` to `rust-lang/rust`, so this can be merged now.

Context: https://github.com/rust-lang/infra-team/issues/188

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Dev
+
+### Changed
+- Unrolled CI builds have moved from the `rust-lang-ci/rust` repository to the `rust-lang/rust` repository.
+
 ## v0.6.9
 
 ### Added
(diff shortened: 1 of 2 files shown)


lqd:
CI commits don't usually last that long before being GCed by github, but will this impact bisecting through recent rollups that were still using `rust-lang-ci`?

Kobzol:
I guess it could, yeah. It's just the URL that would be wrong. We could either provide both links, or test if the link is 404 (?), or add some check based on the date of the commit.

Kobzol:
Added a changelog entry and also the second URL in the printed log.

Kobzol:
It should essentially be a regular try build commit, yeah.

PR #388, opened 2025-07-20
Clarify documentation around usage of merge-base

Kobzol:
https://github.com/rust-lang/cargo-bisect-rustc/pull/387#discussion_r2217507720


src/github.rs
```diff
@@ -82,6 +82,9 @@ fn headers() -> Result<HeaderMap, InvalidHeaderValue> {
     Ok(headers)
 }
 
+/// Lookup information about the given commit reference from GitHub.
+/// If it is a commit forked from the "mainline" git history (e.g. a beta/stable tag),
```
VorpalBlade, line 86: When I tried, I couldn't find a way to specify beta. --start=1.89.0 didn't work. Nor did --start=beta. So are you sure about the "beta tag" bit?

Kobzol: Hmm, good point, beta releases don't actually have tags in git. Reworded.

ehuss (approved):
Thanks!

Issue #205, opened 2022-08-30
feature: show output of tests


steffahn:
I often use the `--prompt` option and then hold down the return key for a while because the (pre-selected) automatic analysis based on `--regress` or a script ***is correct***, yet I want to see the actual output to double check that e.g. there’s no other ICE or error sneaking in or whatnot. I feel like there should be a flag for showing the output without such workarounds. (Unless I’m missing the fact that there might already be such an option.)

ehuss:
`-vv` should show the output. Does that work for you?

steffahn:
I’ll test that shortly…

steffahn:
Yes it does work, thanks :-) But that should probably be better documented. (Also I didn’t quite see yet what `-v` does compared to no verbosity level at all.)

ehuss:
Another minimum thing would be to always show the output if the "start" fails. That often happens when either you pick the wrong start point, or your test itself is broken. Either way, the output is very important.

Unfortunately the logic around capturing the output is a little complicated.

RalfJung:
> `-vv` should show the output. Does that work for you?

Oh wow, that's a well-hidden feature. I tried `-v` to no avail. I didn't expect that repeating the `-v` would make any difference...

PR #437, opened 2026-08-10
Add verbiage to --help

apiraino:
From [comment](https://github.com/rust-lang/cargo-bisect-rustc/issues/205#issuecomment-5236529424)

Now the help mentions that `verbose` can be double  :)

```
-v, --verbose...
          Multiple -v options increase verbosity, including the output of the bisection command. The maximum is 2.
```

@RalfJung does it look better?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -146,7 +146,7 @@ struct Opts {
 
     #[arg(
         short, long = "verbose",
-        help = "Multiple -v options increase verbosity, including the output of the bisection command. The maximum is 2.",
+        help = "Multiple -v options increase verbosity. At level 2, the output of the bisection command is shown.",
         action = ArgAction::Count)]
     verbosity: u8,
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -144,7 +144,10 @@ struct Opts {
     )]
     timeout: Option<usize>,
 
-    #[arg(short, long = "verbose", action = ArgAction::Count)]
+    #[arg(
+        short, long = "verbose",
+        help = "Multiple -v options increase verbosity, including the output of the bisection command. The maximum is 2.",
+        action = ArgAction::Count)]
     verbosity: u8,
 
     #[arg(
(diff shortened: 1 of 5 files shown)


RalfJung:
Thanks! Maybe also mention that the output of the checked program (or whatever term you are using for this) is shown with level 2?

apiraino:
I am not sure what you were looking for, but I hope the updated text looks now better.

RalfJung:
I can confirm that `-vv` does print the program output, in the case of what I was bisecting that was
```
I will run `"rustup" "component" "add" "rust-src"` to install the `rust-src` component for the selected toolchain. Proceed? [Y/n] error: toolchain 'bisector-nightly-2026-08-04-x86_64-unknown-linux-gnu' does not support components
fatal error: failed to install the `rust-src` component for the selected toolchain
```
That can be extremely important information to figure out why the bisect does not behave as expected.

src/main.rs
```diff
@@ -144,7 +144,10 @@ struct Opts {
     )]
     timeout: Option<usize>,
 
-    #[arg(short, long = "verbose", action = ArgAction::Count)]
+    #[arg(
+        short, long = "verbose",
+        help = "Multiple -v options increase verbosity, including the output of the bisection command. The maximum is 2.",
```
RalfJung:
```suggestion
        help = "Multiple -v options increase verbosity. At level 2, the output of the bisection command is shown.",
```

apiraino:
thanks Ralf. I've added a small additional tip also for thr `--script` parameter

RalfJung:
That's potentially confusing since `-vv` shows the output also for the cargo command without `--script` (I did not use `--script` when I ran into this).

apiraino:
ok, reverted to only changing the inline help for `-v` / `-vv` parameters.

RalfJung:
Thanks :)

apiraino:
I don't know who is really in charge for a review here, I'll merge my own PR

